PR #8, opened 2015-03-06
Update rmw_take signature

esteve:
Updates caused by ros2/rmw#4

Connects to ros2/rmw_connext#3

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -338,7 +338,7 @@ rmw_destroy_subscription(rmw_subscription_t * subscription)
 }
 
 rmw_ret_t
-rmw_take(const rmw_subscription_t * subscription, void * ros_message)
+rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * taken)
 {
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -352,7 +352,9 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message)
   const message_type_support_callbacks_t * callbacks = \
     subscriber_info->callbacks;
 
-  return callbacks->take(topic_reader, ros_message);
+  *taken = callbacks->take(topic_reader, ros_message);
```
dirk-thomas: We should either pass a reference instead of a pointer or check for null before using it. I would prefer the first option.

esteve: Unfortunately we can't use a reference, this needs to be C. I'll update this to check for null.

dirk-thomas: We should rename the file then.

esteve: The code in this file is C++, but the function signatures are in C to comply with the API in `rmw.h`

tfoote:
+1

PR #14, opened 2015-03-19
Use opensplice_cpp_identifier as a symbol identifier

esteve:
Connects ros2/ros2#18

### DIFF
--- a/rmw_opensplice_cpp/CMakeLists.txt
+++ b/rmw_opensplice_cpp/CMakeLists.txt
@@ -41,6 +41,11 @@ endif()
 ament_package()
 
 install(
+  DIRECTORY include/
+  DESTINATION include
+)
+
+install(
   TARGETS rmw_opensplice_cpp
   LIBRARY DESTINATION lib
   ARCHIVE DESTINATION lib
--- a/rmw_opensplice_cpp/CMakeLists.txt
+++ b/rmw_opensplice_cpp/CMakeLists.txt
@@ -18,7 +18,7 @@ find_package(rosidl_typesupport_opensplice_cpp REQUIRED)
 
 ament_export_definitions(${OpenSplice_DEFINITIONS})
 ament_export_dependencies(rmw rosidl_generator_cpp rosidl_typesupport_opensplice_cpp)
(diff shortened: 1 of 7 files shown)


esteve:
Reviewable but not mergeable

esteve:
@dirk-thomas @tfoote @wjwwood ping

rmw_opensplice_cpp/CMakeLists.txt
```diff
@@ -41,6 +41,11 @@ endif()
 ament_package()
 
 install(
+  DIRECTORY include/
+  DESTINATION include
+)
```
dirk-thomas: Does it also need to export the include directories?

esteve: The include directories are needed for `rclcpp`

dirk-thomas: Then I should also list `include` as an argument for `ament_export_include_directories(..)`.

esteve: I think I don't really understand what `ament_export_include_directories` does. In this case, I'm exporting `include/${PROJECT_NAME}/impl` because I need `rclcpp` to be able to include `rmw/get_type_support_handle.h`, but I don't need any other package to be able to include the full include path (i.e. `rmw_opensplice_cpp/impl/rmw/get_type_support_handle.h`).

However I still need to install the `include/` directory, or the downstream packages can't find the header files. Do I have to add `include` to `ament_export_include_directories` even if I don't want it to be exported?

dirk-thomas: `ament_export_include_directories(..)` will make values available to downstream packages in the variable `pkgname_INCLUDE_DIRS`. A relative path as an argument will be relative to the `CMAKE_INSTALL_PREFIX` and should usually be `include`. An absolute path is well absolute and should not point to the source / build space.

I don't fully understand what you want to install where in this case so I can't give a specific answer what to do here.

dirk-thomas: After talking about it the current state looks good as-is.

dirk-thomas:
lgtm

Issue #10, opened 2015-03-07
Add support for service for Opensplice


esteve:
Acceptance criteria:

- [ ] Client and server examples work

Does not need to be interoperable with RTI Connext for now.

PR #15, opened 2015-03-19
Opensplice services

esteve:
Adds infrastructure for services in OpenSplice

This is not ready for review, I just created this PR to connect it with the Waffle card

Connects to #10


esteve:
@dirk-thomas @tfoote @wjwwood this is now ready for review

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -499,6 +532,56 @@ rmw_wait(rmw_subscriptions_t * subscriptions,
...
+    unsigned long j = 0;
+    for (; j < active_conditions.length(); ++j)
+    {
+      if (active_conditions[j] == condition)
+      {
+        break;
+      }
+    }
+    // if service condition is not found in the active set
+    // reset the subscriber handle
```
dirk-thomas: "reset the service handle"

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -499,6 +532,56 @@ rmw_wait(rmw_subscriptions_t * subscriptions,
...
+    unsigned long j = 0;
+    for (; j < active_conditions.length(); ++j)
+    {
+      if (active_conditions[j] == condition)
+      {
+        break;
+      }
+    }
+    // if client condition is not found in the active set
+    // reset the subscriber handle
```
dirk-thomas: "reset the client handle"

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -508,34 +591,94 @@ rmw_create_client(const rmw_node_t * node,
...
+  if (client->implementation_identifier != opensplice_cpp_identifier)
+  {
+    rmw_set_error_string("client handle not from this implementation");
+    // printf("but from: %s\n", client->implementation_identifier);
+    return RMW_RET_ERROR;
+  }
+
+  OpenSpliceStaticClientInfo * client_info = \
+    static_cast<OpenSpliceStaticClientInfo*>(client->data);
+
```
dirk-thomas: I would suggest to remove these empty lines (similar to the block above).

rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
```diff
@@ -45,6 +45,10 @@ def main(argv=sys.argv[1:]):
         help='The location of the template files')
     args = parser.parse_args(argv)
 
+    print("BAR FOOOOOOO: ROS FILES")
+    print(', '.join(args.dds_interface_files))
+    print(repr(sys.argv))
+
```
dirk-thomas: Remove debug output.

esteve:
@dirk-thomas @tfoote @wjwwood Sorry, this is not ready for review, I wrote on the wrong PR

rosidl_typesupport_opensplice_cpp/resource/msg_TypeSupport.h.template
```diff
@@ -20,6 +20,15 @@
 #include "rosidl_generator_cpp/MessageTypeSupport.h"
 #include "rosidl_typesupport_opensplice_cpp/MessageTypeSupport.h"
 
+namespace rmw_opensplice_cpp
```
dirk-thomas: Why is the `rosidl` package defining a `rmw` namespace?

esteve: This is not ready for review yet, I wrote the "ping" comment on the wrong PR.

wjwwood: I think this is a forward declaration.

esteve:
@dirk-thomas @tfoote @wjwwood this should be now ready for review, examples work.

esteve:
This PR depends on ros2/rosidl_dds#10

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -508,34 +591,89 @@ rmw_create_client(const rmw_node_t * node,
...
+  client->data = client_info;
+  return client;
 }
 
 rmw_ret_t
 rmw_destroy_client(rmw_client_t * client)
 {
-  return RMW_RET_OK;
+  if (client) {
+    // TODO(esteve): de-allocate Requester and response DataReader
```
tfoote: Is this done? or does it need to be done?

esteve: It's not done, I'll add a new function to the callbacks structure to destroy the Requester in a typesafe way.

tfoote:
lgtm

dirk-thomas:
lgtm

esteve:
@dirk-thomas I just pushed a change so that we only use a single APPEND call for adding all the dependency files.

dirk-thomas:
Awesome!

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -6,7 +6,6 @@
 #include <rmw/allocators.h>
 #include <rmw/error_handling.h>
 #include <rmw/rmw.h>
-#include <rosidl_generator_c/message_type_support.h>
```
wjwwood: Why remove this? It provides the symbol `rosidl_message_type_support_t` which is used in this file.

esteve: I just readded this. I also added `#include <rosidl_generator_c/service_type_support.h>` for `rosidl_service_type_support_t`

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
```diff
@@ -0,0 +1,126 @@
+#ifndef __rosidl_typesupport_opensplice_cpp__Requester__h__
```
wjwwood: Two things:

- Can you put in a copyright/license header here?
- Even though it's currently very inconsistent, I believe we've decided to do all lowercase with underscores for file names, so this would be `requester.hpp` rather than `Requester.hpp` (the include guard should be updated to match). Later I'll make a pass to make everything consistent, but we should try to match the expected style on new stuff.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
```diff
@@ -0,0 +1,126 @@
...
+#define __rosidl_typesupport_opensplice_cpp__Requester__h__
+
+#include <atomic>
+#include <random>
+#include <limits>
+#include <ccpp_dds_dcps.h>
+#include <u_entity.h>
```
wjwwood: Can you separate the system headers from the external library headers with a new line?

Also, where does `u_entity.h` come from? Is that a dds header?

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+namespace rosidl_typesupport_opensplice_cpp
+{
+
+template<typename RequestT, typename ResponseT>
+class Requester
+{
+  public:
+    Requester(DDS::DomainParticipant_ptr participant, const char * service_name,
+              const char * service_type_name) :
```
wjwwood: Should we used `std::string` here or `const char *`? I'm not sure `std::string` is better, I'm just curious to see what others think.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
esteve: It's part of the internal OpenSplice API, I had some code that used internal functions that might uniquely identify `DataWriter`s, but I ended up using random `uint64`s instead.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
esteve: I renamed both to lowercase and added the Apache License header.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
esteve: I'd rather use `std::string` here, but everywhere else `const char *` is being used in the template files.

wjwwood: I'd say change them to `const std::string` unless someone else objects. This will never be exposed to a C interface (it's class after all).

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
```diff
@@ -0,0 +1,141 @@
...
+#define __rosidl_typesupport_opensplice_cpp__requester__h__
+
+#include <atomic>
+#include <limits>
+#include <random>
+
+#include <ccpp_dds_dcps.h>
+
+#include <rosidl_typesupport_opensplice_cpp/message_type_support.h>
+#include <rosidl_typesupport_opensplice_cpp/service_type_support.h>
```
wjwwood: Something else we are pretty inconsistent about, should this be `""` instead of `<>`? These headers are coming from this package, does anyone have an opinion on what our style should be?

tfoote: I believe we should use `"` for this package and `<` for other packages.

esteve: I agree with @tfoote

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
dirk-thomas: I wouldn't start changing it right now. A lot of libraries use `const char *` when possible to avoid the "overhead" of std:string. I think as long as we don't need std:string like functionality we should stick to plain `const char *`.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
wjwwood: Reading up on the difference, it seems that when a `""` search fails it is implicitly converted to `<>` and the search is resumed. On some systems only the `""` style include will search relative to the current file, but that doesn't seem standard. In general relative includes are not recommended anyways.

Basically it seems to not matter, so we can use (abuse?) the lack of distinction to suggest the origin of the header w.r.t. our packaging structure, as @tfoote suggested.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/Requester.hpp
esteve: The issue here is that there is some overhead to using `const char *`. These variables and arguments are converted between `std::string` and `const char *`, back and forth. I'd prefer passing `std::string`s around and only convert to `const char *` when they need to be passed to a C API.

wjwwood: The class is immediately storing the incoming `const char *` in a `std::string` class member variable. So in this case (AFAIU) changing the parameter will not have any effect on performance.

But in general I would prefer the use of `std::string` simply to be idiomatic to C++ and only in exceptional cases where a performance gain can be made (and needs to be made) use the `const char *`. Other reasons to use `const char *` are if you're going to be coming or going through a C API. I guess what I'm getting at is that I would prefer to default to `std::string` unless there is a good reason to not use it, rather than the other way around. Also, a lot of libraries made the decision to avoid the overhead of STL containers before the move semantics and rvalue's were introduced, so it might not be the best reason to adopt the same policy.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+template<typename RequestT, typename ResponseT>
+class Responder
+{
+  public:
+
+    Responder(DDS::DomainParticipant_ptr participant, const char * service_name,
+      const char * service_type_name) : participant_(participant),
+      service_name_(service_name), service_type_name_(service_type_name) {
+      // Create request Publisher and DataWriter
```
wjwwood: Can you put the `{` on a new line aligned with the start of the method declaration? I find it hard to look at this and see where the declaration ends and the definition begins.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
```diff
@@ -0,0 +1,121 @@
...
+      DDS::ReturnCode_t status;
+      DDS::TopicQos default_topic_qos;
+      status = participant->get_default_topic_qos(default_topic_qos);
```
wjwwood: Should we be checking the value of status? (this applies throughout this file)

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+      std::string request_type_name = service_type_name_ + "Request_";
+      std::string request_topic_name = service_name_ + "_Request";
+
+      request_topic_ = participant->create_topic(
```
wjwwood: Inline with my previous comment, is `request_topic_` guaranteed to be valid? Should we check its validity?

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/service_type_support.h
```diff
@@ -7,6 +7,24 @@ typedef struct service_type_support_callbacks_t
 {
   const char * package_name;
   const char * service_name;
+  // Function to create a requester
+  void* (*create_requester)(void * participant, const char * service_name,
+                            void **reader);
+  // Function to create a responder
+  void* (*create_responder)(void * participant, const char * service_name,
+                          void **reader);
```
wjwwood: nitpick: indention isn't aligned

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/service_type_support.h
```diff
@@ -7,6 +7,24 @@ typedef struct service_type_support_callbacks_t
 {
   const char * package_name;
   const char * service_name;
+  // Function to create a requester
+  void* (*create_requester)(void * participant, const char * service_name,
+                            void **reader);
```
wjwwood: Should the newly created requester also be passed as a `**`? It seems odd to me to return one object as the return value and another as a `**` parameter. Usually when there is multiple return values, we either need to encapsulate in a structure or both are returned through `**` parameters and a return code is returned. Any thoughts on this?

rosidl_typesupport_opensplice_cpp/resource/srv_ServiceTypeSupport.cpp.template
```diff
@@ -15,14 +15,25 @@
...
 #include <rosidl_generator_c/visibility_control.h>
 // This is defined in the rosidl_typesupport_opensplice_cpp package and
 // is in the include/rosidl_typesupport_opensplice_cpp/impl folder.
+#include <rosidl_generator_cpp/message_type_support.hpp>
 #include <rosidl_generator_cpp/service_type_support.hpp>
 
 #include "@(spec.pkg_name)/@(spec.srv_name)_Service.h"
 #include "rosidl_typesupport_opensplice_cpp/service_type_support.h"
+#include "rosidl_typesupport_opensplice_cpp/requester.hpp"
+#include "rosidl_typesupport_opensplice_cpp/responder.hpp"
```
wjwwood: We should group the `<>` include and the `""` include separately and alphabetize them. I'd expect something like this:

```c++
 #include <rosidl_generator_c/service_type_support.h>
 #include <rosidl_generator_c/visibility_control.h>
 // This is defined in the rosidl_typesupport_opensplice_cpp package and
 // is in the include/rosidl_typesupport_opensplice_cpp/impl folder.
+#include <rosidl_generator_cpp/message_type_support.hpp>
 #include <rosidl_generator_cpp/service_type_support.hpp>
+#include <rmw/rmw.h>
 
 #include "@(spec.pkg_name)/@(spec.srv_name)_Service.h"
+#include "@(spec.pkg_name)/dds_opensplice/@(spec.srv_name)Request_TypeSupport.h"
+#include "@(spec.pkg_name)/dds_opensplice/@(spec.srv_name)Response_TypeSupport.h"
+#include "@(spec.pkg_name)/dds_opensplice/ccpp_@(spec.srv_name)Request_.h"
+#include "@(spec.pkg_name)/dds_opensplice/ccpp_@(spec.srv_name)Response_.h"
+#include "@(spec.pkg_name)/dds_opensplice/ccpp_Sample@(spec.srv_name)Request_.h"
+#include "@(spec.pkg_name)/dds_opensplice/ccpp_Sample@(spec.srv_name)Response_.h"
 #include "rosidl_typesupport_opensplice_cpp/service_type_support.h"
+#include "rosidl_typesupport_opensplice_cpp/requester.hpp"
+#include "rosidl_typesupport_opensplice_cpp/responder.hpp"
```

Also the `#include "@(spec.pkg_name)/@(spec.srv_name)_Service.h"` occurs twice.

rosidl_typesupport_opensplice_cpp/resource/srv_ServiceTypeSupport.cpp.template
```diff
@@ -31,20 +42,271 @@ class DataReader;
...
+
+void destroy_requester__@(spec.srv_name)(void * untyped_requester)
+{
+  rosidl_typesupport_opensplice_cpp::Requester<@(spec.pkg_name)::dds_::@(spec.srv_name)Request_, @(spec.pkg_name)::dds_::@(spec.srv_name)Response_> * requester = reinterpret_cast<rosidl_typesupport_opensplice_cpp::Requester<@(spec.pkg_name)::dds_::@(spec.srv_name)Request_, @(spec.pkg_name)::dds_::@(spec.srv_name)Response_> *>(untyped_requester);
+  delete requester;
+}
+
+void destroy_responder__@(spec.srv_name)(void * untyped_responder)
+{
+  rosidl_typesupport_opensplice_cpp::Responder<@(spec.pkg_name)::dds_::@(spec.srv_name)Request_, @(spec.pkg_name)::dds_::@(spec.srv_name)Response_> * responder = reinterpret_cast<rosidl_typesupport_opensplice_cpp::Responder<@(spec.pkg_name)::dds_::@(spec.srv_name)Request_, @(spec.pkg_name)::dds_::@(spec.srv_name)Response_> *>(untyped_responder);
```
wjwwood: These lines are super long, it's really painful to read them in this diff. Can you try to shorten them?

rosidl_typesupport_opensplice_cpp/resource/srv_ServiceTypeSupport.cpp.template
esteve: Done.

rosidl_typesupport_opensplice_cpp/resource/srv_ServiceTypeSupport.cpp.template
esteve: Done.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
wjwwood: It's probably a lot of work to check the return values "correctly" or at least to do the right thing when it's not what we expect. So I would be ok with `TODO`'s for each of them. My main concern is that we don't do them now and never notice that they're not being check in the future.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/service_type_support.h
esteve: Done, requesters and responders are now returned via parameters.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/service_type_support.h
esteve: Done.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
esteve: Done.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
esteve: Yes, though I'm not sure what'd be the best way to do it here:

* Raise an exception and let the code in the template call `rmw_set_error_string()`
* Call `rmw_set_error_string` from here

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
esteve: There are many cases where `request_topic_` could be `NULL`, for example if the type of the topic is not registered.

esteve: So, yes, this should be checked.

esteve: Done. Added `TODO`s for each newly created DDS entity.

wjwwood:
Looks good to me.

PR #17, opened 2015-03-25
Fix get type support handle

wjwwood:
This pull request updates the layout of the opensplice packages to work with https://github.com/ros2/rosidl/pull/9. It also changes the way the opensplice env hook works so that it supports both our build of opensplice and the one provided by PrismTech, which is something I had to fix during my testing and is isolated to this commit: f2c08be.

Connects to ros2/ros2#26


tfoote:
+1

esteve:
+1

dirk-thomas:
The same comments from ros2/rmw_connext#11 also apply here.

PR #19, opened 2015-03-30
Fix compilation on Windows because of min/max macros

### LINKED ISSUE #18: VS2015 can't compile OpenSplice services
Unfortunately `<windows.h>` defines `max` and `min` as macros, so `std::numeric_limits<T>::min` and `std::numeric_limits<T>::max` don't work properly. We use those two functions to generate unique client ids in `requester.hpp`

esteve:
Avoid macro expansion because `<windows.h>` defines `max` and `min` as macros, which conflict with `std::numeric_limits<T>::max()` and `std::numeric_limits<T>::min()`

Fixes #18

@dirk-thomas @tfoote @wjwwood


dirk-thomas:
+1

wjwwood:
+1

PR #21, opened 2015-03-31
remove dhcpisocpp from LIBRARIES since we can only link against one

dirk-thomas:
Currently we only use dhcpsacpp.

Connects to ros2/rclcpp#4.


esteve:
+1

wjwwood:
+1, it will be `dcpsisocpp.lib` or `dcpsisocpp.dll` on Windows, but it looks like you have that case covered as well.

tfoote:
+1

PR #22, opened 2015-04-01
quote paths which might have whitespace

tfoote:
Connects to ros2/ros2#20

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/__init__.py
+++ b/rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/__init__.py
@@ -30,14 +30,15 @@ def generate_dds_opensplice_cpp(
 
         cmd = [idl_pp]
         for include_dir in include_dirs:
-            cmd += ['-I', include_dir]
+            cmd += ['-I', '"%s"' % include_dir]
         cmd += [
             '-S',
             '-l', 'cpp',
             '-o', 'dds-types',
-            '-d', output_dir,
-            idl_file
+            '-d', '"%s"' % output_dir,
+            '"%s"' % idl_file
         ]
+        print("CMD IS %s" % cmd)
         subprocess.check_call(cmd)
 
     return 0


rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/__init__.py
```diff
@@ -30,14 +30,15 @@ def generate_dds_opensplice_cpp(
 
         cmd = [idl_pp]
         for include_dir in include_dirs:
-            cmd += ['-I', include_dir]
+            cmd += ['-I', '"%s"' % include_dir]
```
wjwwood: Just out of curiosity did you try `cmd += ['-I"%s" % include_dir']` or even `cmd += ['"-I%s" % include_dir']`?

dirk-thomas:
Why is this necessary at all? Python subprocess already gets the command in a list where each argument is separated from each other.

rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/__init__.py
tfoote: I did not try those. I kept the resultant syntax the same with the space betwen -I and the value.

tfoote:
I don't think this is explicitly necessary for this program. However it seems to pass through the values as passed to the child process a little better. I think the child process is stringifying the arguments and then string processing them. 

Without this you get the following errors; (note CMD IS is a debugging output I added)
```
CMD IS ['/usr/bin/idlpp', '-I', '/home/tfoote/work/r2/ros2 ws/build/simple_msgs/rosidl_generator_dds_idl', '-S', '-l', 'cpp', '-o', 'dds-types', '-d', '/home/tfoote/work/r2/ros2 ws/build/simple_msgs/rosidl_typesupport_opensplice_cpp/simple_msgs/dds_opensplice', '/home/tfoote/work/r2/ros2 ws/build/simple_msgs/rosidl_generator_dds_idl/simple_msgs/dds_opensplice/Uint32_.idl']
Failed to open IDL file: ws/build/simple_msgs/rosidl_generator_dds_idl
Failed to open IDL file: ws/build/simple_msgs/rosidl_generator_dds_idl
simple_msgs /home/tfoote/work/r2/ros2 ws/src/ros2/examples/simple_msgs/msg/Time.msg
```

We could close this as won't fix since it's only a parital solution to the real upstream issue. https://github.com/PrismTech/opensplice/issues/15

rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/__init__.py
dirk-thomas: We should try of the tool handles the cases @wjwwood suggested better.

tfoote:
We should probably hold this until we get a fix from upstream. It shouldn't be necessary, and is only a partial fix.

dirk-thomas:
Agreed.

esteve:
+1

dirk-thomas:
Closing in favor of patching upstream: ros2/ros2#35

PR #23, opened 2015-04-01
Use plain pointer types instead of _ptr

### LINKED ISSUE #20: OpenSplice examples crash on Windows
All examples abort when run on Windows, seems to be related to the use of _ptr vs * for pointers.

esteve:
This PR replaces `_ptr` types with plain pointers.

Fixes #20 

@dirk-thomas @tfoote @wjwwood


dirk-thomas:
lgtm

It would be good to get to an understanding why it doesn't work. But there for sure more important tasks right now.

wjwwood:
+1

PR #24, opened 2015-04-07
code style only

dirk-thomas:
Connects to ros2/ros2#21.

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -33,25 +33,27 @@ RMW_PUBLIC const char * opensplice_cpp_identifier = "opensplice_static";
 
 struct OpenSpliceStaticPublisherInfo
 {
-  DDS::DataWriter* topic_writer;
+  DDS::DataWriter * topic_writer;
   const message_type_support_callbacks_t * callbacks;
 };
 
 struct OpenSpliceStaticSubscriberInfo
 {
-  DDS::DataReader* topic_reader;
+  DDS::DataReader * topic_reader;
   const message_type_support_callbacks_t * callbacks;
 };
 
-struct OpenSpliceStaticClientInfo {
+struct OpenSpliceStaticClientInfo
+{
   void * requester_;
-  DDS::DataReader* response_datareader_;
+  DDS::DataReader * response_datareader_;
   const service_type_support_callbacks_t * callbacks_;
 };
 
-struct OpenSpliceStaticServiceInfo {
+struct OpenSpliceStaticServiceInfo
+{
   void * responder_;
-  DDS::DataReader* request_datareader_;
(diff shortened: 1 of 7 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -66,11 +68,10 @@ rmw_init()
 {
   DDS::DomainParticipantFactory_var dp_factory = \
     DDS::DomainParticipantFactory::get_instance();
-  if (!dp_factory)
-  {
+  if (!dp_factory) {
```
esteve: I thought braces went to the next line, like with `class`/`struct` declarations (see line 35)

dirk-thomas: The conclusion from the last discussion was that they are on the next line for things like:
* namespace
* class
* struct

but are inline for control elements like:
* if
* for
* while

esteve: Ok, sounds good.

esteve:
+1

rmw_opensplice_cpp/src/functions.cpp
wjwwood: The point that we didn't come to consensus on was the ability to put the `{` on a new line for `if`, `for`, and `while` if the parameters are wrapped for line length. For example:

```c++
if (true) {
  ...
}

if (
  SomeType a_really_long_variable_name_that_must_wrap_a,
  SomeOtherType a_really_long_variable_name_that_must_wrap_b)
{
  ...
}

if (
  SomeType a_really_long_variable_name_that_must_wrap_a,
  SomeOtherType a_really_long_variable_name_that_must_wrap_b
) {
  ...
}
```

My point was that the third instance would be good if you could have a trailing `,` on the last parameter, which would minimize the diff. But since that isn't the case putting the closing `)` on the same line as the last parameter doesn't affect the diff-ability. Also the second if statement aligns better with the multi-line function call style:

```c++
int ret = some_function(
  SomeType a_really_long_variable_name_that_must_wrap_a,
  SomeOtherType a_really_long_variable_name_that_must_wrap_b);
```

Honestly, I got in the habit of always putting the `{` on a new line in all cases, because that's an easy rule to follow, but I don't really care about that. However, I do think that the second `if` statement above is preferable to the third one, but again I'm willing to go with the flow if others disagree.

One option is that we enforce `{` on a new line for the `namespace`, `class`, and `struct` and just don't enforce style on the others, leaving the style up to the developer, since the best style to use seems to depend on the scenario.

dirk-thomas: I would also prefer the third option. I finally found the corresponding uncrustify option: `nl_multi_line_cond`. I have updated the PRs where necessary.

tfoote:
lgtm

PR #25, opened 2015-04-14
readd typesupport check

dirk-thomas:
Connects to ros2/ros2#30.

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/CMakeLists.txt
+++ b/rosidl_typesupport_opensplice_cpp/CMakeLists.txt
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(rosidl_typesupport_opensplice_cpp)
 
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+endif()
+
 find_package(ament_cmake REQUIRED)
 
 find_package(opensplice_cmake_module REQUIRED)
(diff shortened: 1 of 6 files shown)


dirk-thomas:

esteve:
+1

rosidl_typesupport_opensplice_cpp/CMakeLists.txt
```diff
@@ -23,6 +27,17 @@ ament_export_include_directories(include include/${PROJECT_NAME}/impl)
 
 ament_python_install_package(${PROJECT_NAME})
 
+add_library(${PROJECT_NAME} SHARED src/identifier.cpp)
```
wjwwood: Would it be possible to roll this `identifier.cpp` into the shared library created in the `rmw_opensplice_cpp` package?:

https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/CMakeLists.txt#L24

As far as I know, unlike the `rosidl_typesupport_introspection` implementation, this `rosidl_` package is always paired to the matching `rmw_` package. The combination is always `user exec = user object files + librmw_opensplice_cpp.so + libmsg_pkg1_opensplice.so + libmsg_pkgN_opensplice.so` and now we're adding ` + librosidl_typesupport_opensplice_cpp.so` for just the identifier symbol. It would be great to simplify it to roll this symbol into the `rmw_opensplice_cpp` shared library, but I guess there might be use cases where that won't work?

dirk-thomas: I don't think it should be in `rmw_opensplice_cpp`. Moving the definition of the symbol to the rmw package would imply that a message package builds a library which contains a declared symbol which is not defined in any of its dependencies.

When the definition is coming from the rmw package I would call that a dependency for each message package. And we don't want to introduce a dependency in that direction.

wjwwood: I guess that's fine. This solution doesn't leave me feeling particularly good about the overhead and complexity for such a simple check, but I don't see a better way to do it with using a manually maintained implementation id in a header.

> Moving the definition of the symbol to the rmw package would imply that a message package builds a library which contains a declared symbol which is not defined in any of its dependencies.

This is how the `rmw` implementations work, excluding the defaults. `rmw` provides declarations and `rmw_<implementation>` provides the definitions, and the client library package will depend on `rmw` but does not necessarily need to depend on `rmw_<implementation>`. Plus the client libraries use symbols who's implementation exists in the generated message libraries which it does not directly depend on; they're only available when a user depends on both the message library and the client library at the same time. So I don't think it is that odd. But I see the point.

> And we don't want to introduce a dependency in that direction.

I'm not saying we should do it, but you should really back up assertions like this. Though it may be clear to you, it isn't clear to me why this is true.

wjwwood:
+1 for merging.

tfoote:
+1

PR #31, opened 2015-05-19
use queue size to ensure history depth

dirk-thomas:
Connects to ros2/ros2#42

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -180,16 +180,23 @@ rmw_create_publisher(
     return nullptr;
   }
 
-  DDS::DataWriterQos default_datawriter_qos;
-  status = dds_publisher->get_default_datawriter_qos(default_datawriter_qos);
+  DDS::DataWriterQos datawriter_qos;
+  status = dds_publisher->get_default_datawriter_qos(datawriter_qos);
   if (status != DDS::RETCODE_OK) {
     rmw_set_error_string("failed to get default datawriter qos");
     // printf("get_default_datawriter_qos() failed. Status = %d\n", status);
     return nullptr;
   }
 
+  // ensure the history depth is at least the requested queue size
+  if (datawriter_qos.history.kind == DDS::KEEP_LAST_HISTORY_QOS &&
+    datawriter_qos.history.depth < queue_size
+  ) {
+    datawriter_qos.history.depth = queue_size;
+  }
+
   DDS::DataWriter * topic_writer = dds_publisher->create_datawriter(
-    topic, default_datawriter_qos, NULL, DDS::STATUS_MASK_NONE);
+    topic, datawriter_qos, NULL, DDS::STATUS_MASK_NONE);
   if (!topic_writer) {
     rmw_set_error_string("failed to create datawriter");
     return nullptr;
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -180,16 +180,23 @@ rmw_create_publisher(
...
+  DDS::DataWriterQos datawriter_qos;
+  status = dds_publisher->get_default_datawriter_qos(datawriter_qos);
   if (status != DDS::RETCODE_OK) {
     rmw_set_error_string("failed to get default datawriter qos");
     // printf("get_default_datawriter_qos() failed. Status = %d\n", status);
     return nullptr;
   }
 
+  // ensure the history depth is at least the requested queue size
+  if (datawriter_qos.history.kind == DDS::KEEP_LAST_HISTORY_QOS &&
```
wjwwood: Why check if the history kind is correct? Why not always set the `history.kind` to that and always set the `history.depth` to `queue_size`?

dirk-thomas: If the user has already configured a "better" history (e.g. keep all or a larger number) then the code should not *reduce* the history depth.

wjwwood: When would the user have had the chance to do that? The publisher was just created in this function. At best they could do it after the publisher is returned from this function.

dirk-thomas: It is always possible to configure QoS parameters externally, e.g. via XML files.

wjwwood: My concern is that if the default changes history kind then the queue size is not even set. Also, if I pass a queue size into this create function I would expect the queue is set to exactly what I asked, regardless of the default history settings.

wjwwood: If we want to be able to fall back to system defaults, then we need a "create default publisher" option. I think it is the wrong behavior for the function to receive a `queue_size` and then potentially not use it.

dirk-thomas: Always setting the history kind and depth here would prevent using vendor specific configuration options completely.

The history kind only support *ALL* and *DEPTH N*. If the default has been set to *ALL* the function doesn't reduce the history. If it is set to a specific *DEPTH* it ensures that it has at least the requested queue size. But if the default has been changed to a larger value it is not reduced.

Imo if the user changes the default configuration he is doing that for a specific reason and I would not overwrite that choice in our code.

wjwwood: The ability to do that is an implementation detail, this function must behave consistently based on the input. I think that if I design my node to explicitly need a queue size of 1 and you set the global to keep all that will not work. If you want to use the defaults then we need a way to specify "use default", which is how it is modeled in DDS. We cannot ask users to always specify a queue_size and then potentially ignore it.

tfoote:
+1

rmw_opensplice_cpp/src/functions.cpp
dirk-thomas: I have created ros2/rmw#13 to possibly extend the rmw interface to support both cases. Until then we will keep the behavior implemented in this PR.

wjwwood:
+1

PR #36, opened 2015-06-09
support OpenSplice 6.5

dirk-thomas:

### DIFF
--- a/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
+++ b/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
@@ -64,7 +64,6 @@ if(NOT "${_ospl_home} " STREQUAL " ")
     "${_ospl_home}/include/dcps/C++/SACPP")
   set(OpenSplice_LIBRARIES
     "cmxml"
-    "commonserv"
     "dcpsgapi"
     "dcpssac"
     "ddsconfparser"
--- a/rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/rosidl_generator_dds_idl_extension.py
+++ b/rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/rosidl_generator_dds_idl_extension.py
@@ -30,10 +30,7 @@ def get_include_directives(spec, subfolders):
     # prepend directives to include opensplice specific idl file
     for field in spec.fields:
         if field.type.pkg_name == 'builtin_msgs':
-            includes[:0] = [
-                '#ifndef OSPL_IDL_COMPILER',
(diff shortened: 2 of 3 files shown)


rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/rosidl_generator_dds_idl_extension.py
```diff
@@ -30,10 +30,7 @@ def get_include_directives(spec, subfolders):
     # prepend directives to include opensplice specific idl file
     for field in spec.fields:
         if field.type.pkg_name == 'builtin_msgs':
-            includes[:0] = [
-                '#ifndef OSPL_IDL_COMPILER',
-                '#include <dds_dcps.idl>',
-                '#endif']
+            includes.insert(0, '#include <dds_dcps.idl>')
```
wjwwood: What's the idea with this change? Why do you no longer need to check if that is defined?

opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
```diff
@@ -64,7 +64,6 @@ if(NOT "${_ospl_home} " STREQUAL " ")
     "${_ospl_home}/include/dcps/C++/SACPP")
   set(OpenSplice_LIBRARIES
     "cmxml"
-    "commonserv"
```
wjwwood: Have you tried this with older versions of OpenSplice? Do we care about that?

rosidl_typesupport_opensplice_cpp/rosidl_typesupport_opensplice_cpp/rosidl_generator_dds_idl_extension.py
dirk-thomas: The usage of `--dds-types` is deprecated and made the DDS builtin types magically available before.

The `ifndef` was a workaround for `idlpp` not finding the `dds_dcps.idl` file. But when correctly adding the template path as an include directory that is not necessary anymore.

wjwwood: Ok, sounds good to me.

opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
dirk-thomas: Of course it still works with the current LGPL version of OpenSplice (6.4).

Afaik we don't use `commonserv`.

wjwwood:
+1

tfoote:
+1

esteve:
+1

PR #38, opened 2015-06-11
Fix for the idl error

wjwwood:
This is a fix for the IDL error: https://github.com/ros2/rcl_interfaces/issues/2

Basically it looks like when the argument length for a custom command approaches 8192 characters long, MSBuild screws up the splicing and re-concatenating of the arguments. Resulting in a single missing character.

This pull request works around this by dumping the largest portion of the arguments into a file in the build space and then passes the path to that file to the python script which reads it and then generates a list of files from the content. Apparently this is referred to as a "response file" in Windows lingo and apparently it is done automatically for extremely long compile commands...

Connects to ros2/rcl_interfaces#2

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
+++ b/rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
@@ -1,11 +1,21 @@
 #!/usr/bin/env python3
 
 import argparse
+import os
 import sys
 
 from rosidl_parser import UnknownMessageType
 from rosidl_typesupport_opensplice_cpp import generate_dds_opensplice_cpp
 from rosidl_typesupport_opensplice_cpp import generate_typesupport_opensplice_cpp
+
+
+def is_valid_file(parser, file_name):
+    if not os.path.exists(file_name):
+        parser.error("File does not exist: '{0}'".format(file_name))
+    file_name_abs = os.path.abspath(file_name)
+    if not os.path.isfile(file_name_abs):
+        parser.error("Path exists but is not a file: '{0}'".format(file_name_abs))
+    return file_name
 
 
 def main(argv=sys.argv[1:]):
@@ -21,9 +31,10 @@ def main(argv=sys.argv[1:]):
         nargs='*',
         help='The ROS interface files')
     parser.add_argument(
-        '--dds-interface-files',
-        nargs='*',
-        help='The DDS interface files')
+        '--dds-interface-files-file',
+        required=True, metavar='FILE',
+        type=lambda x: is_valid_file(parser, x),
+        help='The file containing a list of the DDS interface files')
     parser.add_argument(
         '--dds-interface-base-path',
         required=True,
@@ -59,9 +70,12 @@ def main(argv=sys.argv[1:]):
         return 1
     if rc:
         return rc
+    dds_interface_files = []
+    with open(args.dds_interface_files_file, 'r') as f:
+        dds_interface_files = [l for l in f.read().splitlines() if l]
     return generate_dds_opensplice_cpp(
         args.pkg_name,
-        args.dds_interface_files,
+        dds_interface_files,
         args.dds_interface_base_path,
         args.deps,
         args.output_dir,
--- a/rosidl_typesupport_opensplice_cpp/cmake/rosidl_typesupport_opensplice_cpp_generate_interfaces.cmake
+++ b/rosidl_typesupport_opensplice_cpp/cmake/rosidl_typesupport_opensplice_cpp_generate_interfaces.cmake
@@ -119,12 +119,16 @@ foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
   endforeach()
 endforeach()
 
+set(_dds_idl_files_file "${_output_path}/dds_idl_files.txt")
+string(REPLACE ";" "\n" _dds_idl_files_lines "${_dds_idl_files}")
+file(WRITE "${_dds_idl_files_file}" ${_dds_idl_files_lines})
+
 add_custom_command(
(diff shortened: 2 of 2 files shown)


wjwwood:
@dirk-thomas @esteve @tfoote @jacquelinekay for review.

I suppose we may end up needing to do this in other places as well.

wjwwood:
Windows builds with this branch: http://54.183.26.131:8080/job/ros2_batch_ci_windows/283/

rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
```diff
@@ -59,9 +60,18 @@ def main(argv=sys.argv[1:]):
         return 1
     if rc:
         return rc
+    dds_interface_files = []
+    if not os.path.exists(args.dds_interface_files_file):
+        print("File does not exist: {0}".format(args.dds_interface_files_file), file=sys.stderr)
+        return 1
```
dirk-thomas: Should we let `argparse` only accept a parameter which is an existing file instead?

rosidl_typesupport_opensplice_cpp/cmake/rosidl_typesupport_opensplice_cpp_generate_interfaces.cmake
```diff
@@ -119,12 +119,16 @@ foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
   endforeach()
 endforeach()
 
+set(_dds_idl_files_file "${_output_path}/dds_idl_files.txt")
+string(REPLACE ";" "\n" _dds_idl_files_concat "${_dds_idl_files}")
+file(WRITE "${_dds_idl_files_file}" ${_dds_idl_files_concat})
```
dirk-thomas: Would `_dds_idl_files_lines` be a better name for the variable `_dds_idl_files_concat`?

rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
```diff
@@ -59,9 +60,18 @@ def main(argv=sys.argv[1:]):
...
+    with open(args.dds_interface_files_file, 'r') as f:
+        for line in f.readlines():
+            if not line.strip():
+                continue
+            dds_interface_files.append(line.strip())
```
dirk-thomas: The whole block could be reduced to:

    dds_interface_files = f.read().splitlines()

esteve: Actually, it'd be more like this:

```python
dds_interface_files = [line for line in f.read().splitlines() if line]
```

dirk-thomas: There will never be empty lines, right? So checking for them is not really necessary.

wjwwood: I wouldn't expect there to be one, but I think this is more defensive to check. For example, in case someone was hand crafting a file for testing and left trailing whitespace or an extra line at the end of the file. I'll make it more concise, but I'll leave it defensive.

dirk-thomas: Well if someone would decide to put a file name in there which actually has a leading / trailing space (which is very weird but valid) the "defensive approach would break.

wjwwood: Actually I don't think we can do that since technically trailing whitespace as well as just whitespace are valid filenames, at least on Linux (not that I expect we'll exercise that). I've updated it to be more concise at least.

wjwwood: By "that" I meant the `.strip()`'ing.

rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
wjwwood: Is there an built-in way to check? I think you still have to implement the checking function on your own.

wjwwood:
Comments addressed.

rosidl_typesupport_opensplice_cpp/bin/rosidl_typesupport_opensplice_cpp
```diff
@@ -8,6 +9,15 @@ from rosidl_typesupport_opensplice_cpp import generate_dds_opensplice_cpp
 from rosidl_typesupport_opensplice_cpp import generate_typesupport_opensplice_cpp
 
 
+def is_valid_file(parser, file_name):
+    if not os.path.exists(file_name):
+        parser.error("File does not exist: '{0}'".format(file_name))
+    file_name = os.path.abspath(file_name)
```
dirk-thomas: The validation function might not want to change the argument to an absolute file. That is kind of unexpected.

wjwwood: At the very least it's slightly clearer for the following error message, but I can avoid mutating the input. Either way it shouldn't matter, `abspath` won't work if the relative path wouldn't also work.

wjwwood:
Ok, can we get a +1/-1 from everyone? I'd like to get this merged.

dirk-thomas:
+1

tfoote:
+1

wjwwood:
@esteve I'm going to take your +1 for granted, please comment if you see something after the fact.

esteve:
+1, sorry to be late to the party.

PR #42, opened 2015-06-19
Added basic qos for readers and writers

esteve:
Connects to ros2/ros2#66

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -179,7 +179,7 @@ rmw_create_publisher(
   const rmw_node_t * node,
   const rosidl_message_type_support_t * type_support,
   const char * topic_name,
-  size_t queue_size)
+  const rmw_qos_profile_t & qos_profile)
 {
   if (!node) {
     RMW_SET_ERROR_MSG("node handle is null");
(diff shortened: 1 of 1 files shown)


esteve:
Linux buildfarm:

http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/155/

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -188,6 +189,17 @@ rmw_create_publisher(
...
 
+  switch(qos_policy) {
+    case BEST_EFFORT:
+      datawriter_qos.reliability.kind = DDS::BEST_EFFORT_RELIABILITY_QOS;
+      break;
+    case RELIABLE:
+      datawriter_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;
+      break;
+    default:
+      break;
```
dirk-thomas: This case should throw a `runtime_error` so that it is easy to catch if more qos policies are added but not everywhere handled.

Same below.

esteve:
Linux buildfarm:

http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/156/

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -188,6 +189,17 @@ rmw_create_publisher(
...
 
+  switch (qos_policy) {
+    case RMW_QOS_POLICY_BEST_EFFORT:
+      datawriter_qos.reliability.kind = DDS::BEST_EFFORT_RELIABILITY_QOS;
+      break;
+    case RMW_QOS_POLICY_RELIABLE:
+      datawriter_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;
+      break;
+    default:
+      throw std::runtime_error("Unknown QoS policy");
```
dirk-thomas: It would be good if the error string would include the actual numeric value of `qos_policy`.

Same below.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -270,19 +270,41 @@ rmw_create_publisher(
...
 
+  switch (qos_policy.history) {
+    case RMW_QOS_POLICY_KEEP_LAST_HISTORY:
+      datawriter_qos.history.kind = DDS::KEEP_LAST_HISTORY_QOS;
+      break;
+    case RMW_QOS_POLICY_KEEP_ALL_HISTORY:
+      datawriter_qos.history.kind = DDS::KEEP_ALL_HISTORY_QOS;
+      break;
+    default:
+      throw std::runtime_error("Unknown QoS history policy");
```
dirk-thomas: The rmw implement must never `throw` since it is a C interface.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -270,19 +270,41 @@ rmw_create_publisher(
...
   )
   {
-    if (queue_size > (std::numeric_limits<DDS::Long>::max)()) {
+    if (qos_policy.depth > (std::numeric_limits<DDS::Long>::max)()) {
       RMW_SET_ERROR_MSG(
         "failed to set history depth since the requested queue size exceeds the DDS type");
       goto fail;
     }
-    datawriter_qos.history.depth = static_cast<DDS::Long>(queue_size);
+    datawriter_qos.history.depth = qos_policy.depth;
```
dirk-thomas: Doesn't that reintroduce a compiler warning?

rmw_opensplice_cpp/src/functions.cpp
esteve: I added a throw because of your previous comment :-)

https://github.com/ros2/rmw_connext/pull/42#discussion_r32878674

I'll replace this with an assertion.

dirk-thomas: My previous comment is over a month ago when we did not yet have the error message API in rmw.

Instead of throwing or asserting this should use the following as it is done for all other errors in the rmw implementations:

    RMW_SET_ERROR_MSG("boom");
    goto fail;

esteve: I know it was a month ago, I was hoping my comment wouldn't be taken seriously, that's what the smiley was for... next time I'll use a picture of a kitty.

Anyway, I'll do the goto thing.

rmw_opensplice_cpp/src/functions.cpp
esteve: Good point, will fix it.

dirk-thomas:
+1

jacquelinekay:
Did you run CI for OSX as well as Linux for this branch?

esteve:
@jacquelinekay just started one to make sure:

http://ci.ros2.org/job/ros2_batch_ci_osx/88/

jacquelinekay:
It looks like you need to rebase your "demos" branch so that it will build.

esteve:
@jacquelinekay done, thanks for noticing.

jacquelinekay:
+1

Issue #43, opened 2015-06-23
update implementation to check for error in all cases


dirk-thomas:
Similar to ros2/rmw_connext#43

wjwwood:
This should be resolved by https://github.com/ros2/rmw_opensplice/pull/45

wjwwood:
@dirk-thomas I've re-reviewed this in the light of #45 being merged and I now think it is resolved. I did quite a bit more work after we originally discussed closing this with #45. Please have a look and see what you think.

Note that I created an issue for the destructors of the requester and repliers, but that doesn't really apply to this issue since there are no actions which are not being checked for errors, but instead we're not doing some clean up that we should be doing:

https://github.com/ros2/rmw_opensplice/issues/47

dirk-thomas:
After taking a closer look to the current state I saw a few cases which are not being checked as defensively as in rmw_connext. I will create a PR adding those checks.

dirk-thomas:
Please see #48 for additional checks.

PR #45, opened 2015-06-29
Refactoring of create_* and destroy_* functions

wjwwood:
This is similar to https://github.com/ros2/rmw_connext/pull/45, but is still a work in progress. The same types of changes will be made.

Connects ros2/ros2#69

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -13,32 +13,54 @@
 // limitations under the License.
 
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
 
 #include <ccpp_dds_dcps.h>
 
 #include <rmw/allocators.h>
 #include <rmw/error_handling.h>
+#include <rmw/impl/cpp/macros.hpp>
 #include <rmw/rmw.h>
 #include <rosidl_generator_c/message_type_support.h>
 #include <rosidl_generator_c/service_type_support.h>
+#include <rosidl_typesupport_opensplice_cpp/identifier.hpp>
+#include <rosidl_typesupport_opensplice_cpp/impl/error_checking.hpp>
 #include <rosidl_typesupport_opensplice_cpp/message_type_support.h>
 #include <rosidl_typesupport_opensplice_cpp/service_type_support.h>
+
+#include <rmw/impl/cpp/macros.hpp>
+
+inline std::string
+_create_type_name(
+  const message_type_support_callbacks_t * callbacks,
+  const std::string & sep)
+{
+  return std::string(callbacks->package_name) +
+         "::" + sep + "::dds_::" + callbacks->message_name + "_";
+}
+
+using namespace rosidl_typesupport_opensplice_cpp::impl;
 
 // The extern "C" here enforces that overloading is not used.
 extern "C"
 {
 
-RMW_PUBLIC const char * opensplice_cpp_identifier = "opensplice_static";
+using rosidl_typesupport_opensplice_cpp::typesupport_opensplice_identifier;
+const char * opensplice_cpp_identifier = "opensplice_static";
 
 struct OpenSpliceStaticPublisherInfo
 {
+  DDS::Topic * dds_topic;
+  DDS::Publisher * dds_publisher;
   DDS::DataWriter * topic_writer;
   const message_type_support_callbacks_t * callbacks;
 };
 
 struct OpenSpliceStaticSubscriberInfo
 {
+  DDS::Topic * dds_topic;
+  DDS::Subscriber * dds_subscriber;
   DDS::DataReader * topic_reader;
   const message_type_support_callbacks_t * callbacks;
   bool ignore_local_publications;
@@ -67,8 +89,7 @@ rmw_get_implementation_identifier()
 rmw_ret_t
 rmw_init()
 {
-  DDS::DomainParticipantFactory_var dp_factory = \
-    DDS::DomainParticipantFactory::get_instance();
+  DDS::DomainParticipantFactory_var dp_factory = DDS::DomainParticipantFactory::get_instance();
   if (!dp_factory) {
     rmw_set_error_string("failed to get domain participant factory");
     return RMW_RET_ERROR;
@@ -79,8 +100,7 @@ rmw_init()
 rmw_node_t *
 rmw_create_node(const char * name)
 {
-  DDS::DomainParticipantFactory_var dp_factory = \
-    DDS::DomainParticipantFactory::get_instance();
+  DDS::DomainParticipantFactory_var dp_factory = DDS::DomainParticipantFactory::get_instance();
   if (!dp_factory) {
     rmw_set_error_string("failed to get domain participant factory");
     return nullptr;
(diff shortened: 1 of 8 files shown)


wjwwood:
Ok, I've finished my first past on this pull request too.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -25,6 +25,54 @@
...
+_create_type_name(
+  const message_type_support_callbacks_t * callbacks,
+  const std::string & sep)
+{
+  return std::string(callbacks->package_name) +
+         "::" + sep + "::dds_::" + callbacks->message_name + "_";
+}
+
+// *INDENT-OFF* (prevent uncrustify from using four space indention here)
+#define TRY_PLACEMENT_NEW(Destination, BufferForNew, FailureAction, Type, ...) try { \
```
jacquelinekay: I know this means more refactoring... but... there is code duplication between the macros in this PR and in [rmw_connext 45](https://github.com/ros2/rmw_connext/pull/45/files). Can the macros go in an `rmw` header file?

wjwwood: I explicitly avoided this, since the implementations do not share headers other than from `rmw`. I don't think `rmw` is an appropriate place to put these macros. I'm open to suggestion on how to share them (with `rmw_opensplice` as well).

wjwwood: I should clarify I don't think `rmw` is the right place for this, since `rmw` is currently C only headers, but this would require C++ headers.

I guess it would be ok to add a C++ header to `rmw` under the guise of a "helper header". Maybe like `#include <rmw/impl/cpp/macros.hpp>`.

jacquelinekay: I support this. Doesn't have to happen in this round of PRs, though.

wjwwood:
Should fix https://github.com/ros2/rmw_opensplice/issues/43 too

rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
```diff
@@ -71,13 +73,17 @@ register_type__@(spec.base_type.type)(
...
   @(spec.base_type.pkg_name)::@(subfolder)::dds_::@(spec.base_type.type)_TypeSupport * ros_message_ts =
     new @(spec.base_type.pkg_name)::@(subfolder)::dds_::@(spec.base_type.type)_TypeSupport();
   DDS::ReturnCode_t status = ros_message_ts->register_type(participant, type_name);
   if (status != DDS::RETCODE_OK) {
-    printf("register_type() failed. Status = %d\n", status);
-    throw std::runtime_error("register_type failed");
+    std::stringstream ss;
+    ss << "register_type failed with status: " << status;
+    // TODO(wjwwood): use a custom exception
+    throw std::runtime_error(ss.str());
```
dirk-thomas: The function should not throw but return `bool`.

rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
```diff
@@ -223,13 +235,15 @@ take__@(spec.base_type.type)(
...
+    // TODO(wjwwood): Check if instance handle is valid. Maybe not necessary?
     auto sender_gid = u_instanceHandleToGID(sender_handle);
     auto receiver_gid = u_instanceHandleToGID(receiver_handle);
     ignore_sample = sender_gid.systemId == receiver_gid.systemId;
   }
 
   if (!ignore_sample) {
-    @(spec.base_type.pkg_name)::@(subfolder)::@(spec.base_type.type) & ros_message = *(@(spec.base_type.pkg_name)::@(subfolder)::@(spec.base_type.type) *)untyped_ros_message;
+    @(spec.base_type.pkg_name)::@(subfolder)::@(spec.base_type.type) & ros_message =
+      *(@(spec.base_type.pkg_name)::@(subfolder)::@(spec.base_type.type) *)untyped_ros_message;
```
dirk-thomas: Should check `ros_message` for `null`.

dirk-thomas:
I haven't read the whole patch yet but it does not address #43. It doesn't check for all error cases and especially can't throw.

rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
wjwwood: That isn't sufficient since the way it is setup it does not pass any information back when it returns `false`. The whole function needs to be reimagined. I have TODO's to that effect in similar functions.

wjwwood:
If generated code is in the scope as well, then I'll just move #43 back on the backlog and open a similar one for `rmw_connext`. Because `rmw_connext`'s generated code also throws exceptions, returns false in multiple places within the same function without returning a reason why (usually does fprintf instead), and doesn't properly clean up when a failure occurs mid function.

So far I've avoided major changes to the generated code which are needed to fix the point's you've brought up just now.

dirk-thomas:
For `rmw_connext` this has been addressed in ros2/rmw_connext#43. The generated type support functions are returning `bool` now for that reason. There are cases where `rmw_connext_cpp` throws exceptions but only in cases which are prevented by the way the API is being used.

wjwwood:
If it can throw at all then it should be caught and reported. There should be a consistent way that these functions communicate failure. For the `throw`'s that I *modified*, I catch them and properly report them. For example here:

https://github.com/ros2/rmw_opensplice/pull/45/files#diff-e129723a3d5bd5e2d76c66f2494374f2R609

I chose to modify the `throw`'s and catch them because restructuring the generated functions was going to be so much work. But I guess I'll do it anyways.

rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template
```diff
@@ -105,8 +87,9 @@ public:
       return false;
     }
     if (status != DDS::RETCODE_OK) {
-      printf("take() failed. Status = %d\n", status);
-      throw std::runtime_error("Read failed");
+      std::stringstream ss;
```
jacquelinekay: Including this line causes a compilation errors for the examples on Linux:

```
[ 91%] /home/jackie/code/ros2_ws/build/example_interfaces/rosidl_typesupport_opensplice_cpp/example_interfaces/srv/dds_opensplice/add_two_ints__type_support.cpp: In static member function ‘static bool rosidl_typesupport_opensplice_cpp::TemplateDataReader<example_interfaces::srv::dds_::Sample_AddTwoInts_Request_>::take_sample(DDS::DataReader*, rosidl_typesupport_opensplice_cpp::Sample<example_interfaces::srv::dds_::AddTwoInts_Request_>&)’:
/home/jackie/code/ros2_ws/build/example_interfaces/rosidl_typesupport_opensplice_cpp/example_interfaces/srv/dds_opensplice/add_two_ints__type_support.cpp:79:25: error: aggregate ‘std::stringstream ss’ has incomplete type and cannot be defined
       std::stringstream ss;
                         ^
Building CXX object CMakeFiles/example_interfaces__rosidl_typesupport_opensplice_cpp.dir/rosidl_typesupport_opensplice_cpp/example_interfaces/srv/dds_opensplice/Sample_AddTwoInts_Response_.cpp.o
/home/jackie/code/ros2_ws/build/example_interfaces/rosidl_typesupport_opensplice_cpp/example_interfaces/srv/dds_opensplice/add_two_ints__type_support.cpp: In static member function ‘static bool rosidl_typesupport_opensplice_cpp::TemplateDataReader<example_interfaces::srv::dds_::Sample_AddTwoInts_Response_>::take_sample(DDS::DataReader*, rosidl_typesupport_opensplice_cpp::Sample<example_interfaces::srv::dds_::AddTwoInts_Response_>&)’:
/home/jackie/code/ros2_ws/build/example_interfaces/rosidl_typesupport_opensplice_cpp/example_interfaces/srv/dds_opensplice/add_two_ints__type_support.cpp:165:25: error: aggregate ‘std::stringstream ss’ has incomplete type and cannot be defined
       std::stringstream ss;
```

`#include <sstream>` in this template fixes it.

wjwwood: Good catch, thanks. Fixed in b203d9a

wjwwood:
@dirk-thomas I've update the error handling for pub/sub, but I'm still working on services. Can you have a look at the general approach and give it +1/-1?

wjwwood:
Ok, have a look at 0cd54a3

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -329,105 +488,217 @@ rmw_create_subscription(
...
+  auto result = RMW_RET_OK;
+  OpenSpliceStaticSubscriberInfo * subscription_info =
+    static_cast<OpenSpliceStaticSubscriberInfo *>(subscription->data);
+  if (subscription_info) {
+    DDS::Subscriber * dds_subscriber = subscription_info->dds_subscriber;
+    if (dds_subscriber) {
+      DDS::DataReader * topic_reader = subscription_info->topic_reader;
+      if (topic_reader) {
+        if (topic_reader->delete_contained_entities() != DDS::RETCODE_OK) {
+          rmw_set_error_string("failed to delete contained entiries of datareader");
```
dirk-thomas: Spelling: entiries

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
```diff
@@ -48,98 +53,173 @@ class Requester
...
+      status = participant_->delete_contentfilteredtopic(content_filtered_response_topic_);
+      if (nullptr != impl::check_delete_contentfilteredtopic(status)) {
+        fprintf(stderr, "%s\n", impl::check_delete_contentfilteredtopic(status));
+      }
+    }
+    if (response_datareader_) {
+      // Assumption: subscriber is not null at this point.
+      status = response_subscriber_->delete_datareader(response_datareader_);
+      if (nullptr != impl::check_delete_datareader(status)) {
+        fprintf(stderr, "%s\n", impl::check_delete_datareader(status));
```
dirk-thomas: Should it invoke the function once, check the return value and contionally use the return value for the error message? I can do those changes after merging the PR.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/impl/error_checking.hpp
```diff
@@ -0,0 +1,271 @@
...
+
+#include "builtin_interfaces/msg/duration__struct.hpp"
+
+namespace rosidl_typesupport_opensplice_cpp
+{
+namespace impl
+{
+
+inline const char *
+check_get_default_topic_qos(DDS::ReturnCode_t & status)
```
dirk-thomas: Since the status is a primitive type it should not be passed as a reference.

Same below.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
wjwwood: I thought about doing:

```c++
if (nullptr != (auto error = impl::check_delete_datareader(status))) {
  fprintf(stderr, "%s\n", error);
}
```

But I didn't want to introduce more variables to the local scope. The function is fast and is only called twice in the event of an actual error, so I don't think it matters that much.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/impl/error_checking.hpp
wjwwood: 1b2dd75

rmw_opensplice_cpp/src/functions.cpp
wjwwood: 1b2dd75

dirk-thomas:
+1 for squash and merge.

As a follow up I would replace the usage of `stringstream` for error messages with `fprintf` and change some of the indentation to avoid "aligning".

jacquelinekay:
+1

jacquelinekay:
+q

PR #46, opened 2015-07-13
refactor generators, fix timestamp of generated files

dirk-thomas:
Depends on ros2/rosidl#28.


dirk-thomas:
This fails on a clean workspace - looking into it...

dirk-thomas:
Fixed.

PR #50, opened 2015-07-20
change instance flag when taking samples

dirk-thomas:
~~Fix order of taken samples.~~

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
+++ b/rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
@@ -251,7 +251,7 @@ take__@(spec.base_type.type)(
     1,
     DDS::ANY_SAMPLE_STATE,
     DDS::ANY_VIEW_STATE,
-    DDS::ALIVE_INSTANCE_STATE);
+    DDS::ANY_INSTANCE_STATE);
 
   const char * errs = nullptr;
   bool ignore_sample = false;


wjwwood:
Can you summarize what it was doing before and what's happening now?

dirk-thomas:
Before it was reading `ANY_SAMPLE_STATE` and did not consider if the sample had been read already. So the same sample can be returned over and over again.

Using `ANY_INSTANCE_STATE` is done to make it similar to Connext and not restrict the read sample to alive sources.

wjwwood:
> Before it was reading ANY_SAMPLE_STATE and did not consider if the sample had been read already. So the same sample can be returned over and over again.

So if you are calling take in a multi-threaded scenario you could take the same sample twice? That was not my understanding of how `take` worked. I thought that could only happen if you called `read` and then `take`. Since we're not using `read` I don't think this is necessary. Even if we were using `read`, I think you'd want to leave it as is, otherwise samples that were read with `read` would never be taken.

-0, in this case I think you want the more inclusive selector.

> Using ANY_INSTANCE_STATE is done to make it similar to Connext and not restrict the read sample to alive sources.

+0, It's not clear to me that this is desirable behavior, but it seems like a good idea to take samples even if their datawrite has disposed of them. I guess this could happen when the publisher has a small queue and the subscriber has a larger one.

dirk-thomas:
The demo publishing images at 30fps does not work correctly with the current settings. The video "jumps  back in time" all the time. Something like this sequence: `acdebghif...`.

With this change the problem is fixed. I am happy to hear other proposals how to receive the sample in the correct order. The proposed patch is also in-line with what the two rmw_connext implementations currently do.

wjwwood:
I can't think of how `ANY_SAMPLE_STATE` would affect the ordering. I can imagine that `ANY_INSTANCE_STATE` could affect the delivery order, but I'm not entirely certain by what mechanism it is doing so. Maybe one of those settings is changing some internal behavior which results in ordering the output which is not clear to us.

Did you try explicitly setting the ordered presentation QoS on the data reader? It would be something like this:

```c++
subscriber_qos.presentation.ordered_access = true;
```

In the `create_subscription` function. It seems to me that this is the "correct" way to address ordering issues, but I haven't tried to see if this has any effect on your issue.

dirk-thomas:
I don't think that `ordered_access` can help with this.

When we call `take` twice the API will return *the same* sample when using `ANY_SAMPLE_STATE`. The only way I am aware of is to use the `NOT_READ_SAMPLE_STATE` instead. Because basically the only thing the available samples are different by then is that one has been read before.

How would you propose to take every sample only once while keeping `ANY_SAMPLE_STATE`?

wjwwood:
From the OpenSplice C++ reference (about `take`):

> This operation reads a sequence of Foo samples from the FooDataReader and by doing so, removes the data from the FooDataReader, so it can not be read or taken again. The behaviour is identical to read except for that the samples are removed from the FooDataReader.

Based on that, I don't see how two calls to take could return the same sample. Even if that were the case, that would not result in out of order messages but instead duplicate messages.

dirk-thomas:
I can't deterministically reproduce the problem. I saw it at least once with `ANY_SAMPLE_STATE` too. Therefore I modified this PR to only change the instance state to any.

I have created ros2/rmw_connext#57 to use the same flag for Connext.

jacquelinekay:
To summarize: before this PR, `DataReader` only read samples from alive instances of `DataWriter`. In Dirk's demo, the "live" instance of `DataWriter` writes the messages out of order.

It makes sense that changing `ANY_SAMPLE_STATE` to anything else wouldn't fix the problem, since as William said, this would only affect messages read on `DataReader::read`, and we are calling `take`, which removes the message from the `DataWriter` when successful (according to the OpenSplice docs).

Changing the sample flag from `ALIVE_INSTANCE_STATE` to `ANY_INSTANCE_STATE` causes the samples to come in order (we think).

My current theory is that sample instances have something like a "time to live", and the instance is considered "dead" after the TTL expires. So if the DataReader only reads sample instances that are "alive", it will read the recently written messages by `DataWriter` that haven't yet been dropped. But if the QoS is set to reliable, then the `DataWriter` will make new instances of messages if they haven't been read yet. This is probably why Dirk was seeing messages arrive out of order.

But, `DataWriter` keeps a cache of "dead" sample instances, and when `ANY_INSTANCE_STATE` is set, the reader will accept those sample instances. Rejecting "dead" samples is very important in use cases where the age of messages matters (real-time!) but in the case of a video stream, ordering is far more important. Perhaps we should keep this in mind when we expose QoS parameters in the API.

I haven't confirmed the definition of "alive" or "dead" instances, but I'm convinced by my theory to think this is a good solution for the demo. +1

dirk-thomas:
I don't think the order problem is related to the instance state.

And since the order problem happened with both sample states it is also not address by that.

Currently I don't know how to approach the order problem and I don't know a way to reproduce it either.

wjwwood:
+1, if this change helps then I'm ok with it since, at least imo, it makes more sense as a default take setting anyways.

@dirk-thomas I know you're having trouble reproducing the issue, but did you try the ordered access presentation option? It wasn't clear to me why you think that would not have any affect on the order in which take returns the data? You can also tell DDS to order the presentation by source timestamp, which in my opinion should address this problem.

Here's the snippet from the RTI documentation that lead me to suggest this option in the first place:

http://community.rti.com/rti-doc/45f/ndds.4.5f/doc/html/api_dotnet/structDDS_1_1DestinationOrderQosPolicy.html
> The setting DDS::DestinationOrderQosPolicyKind::BY_SOURCE_TIMESTAMP_DESTINATIONORDER_QOS indicates that (assuming the OWNERSHIP_STRENGTH allows it, within each instance) the source_timestamp of the change shall be used to determine the most recent information. That is, data will be delivered by a DDS::DataReader in the order in which it was sent. If data arrives on the network with a source timestamp that is later than the source timestamp of the last data delivered, the new data will be dropped. This 'by source timestamp' ordering therefore works best when system clocks are relatively synchronized among writing machines.

Now this may have adverse affects on the performance, so this might not be a good solution because of that, but I'm not certain that is the case either.

PR #51, opened 2015-07-21
raise warning level

dirk-thomas:
Connects to ros2/ros2#71

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
+++ b/rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
@@ -114,6 +114,10 @@ register_type__@(spec.base_type.type)(
 void
 convert_ros_message_to_dds(const __ros_msg_type & ros_message, __dds_msg_type & dds_message)
 {
+@[if not spec.fields]@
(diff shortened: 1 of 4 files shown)


rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.template
```diff
@@ -194,7 +198,11 @@ publish__@(spec.base_type.type)(
 }
 
 void
-convert_dds_message_to_ros(const __dds_msg_type & dds_message, __ros_msg_type & ros_message)
+convert_dds_message_to_ros(
+  const __dds_msg_type & @
+@[if not spec.fields]/*@[end if]dds_message@[if not spec.fields]*/@[end if],
+  __ros_msg_type & @
+@[if not spec.fields]/*@[end if]ros_message@[if not spec.fields]*/@[end if])
```
wjwwood: Seems like this should be done the same as the one on line 117.

dirk-thomas: Done: cb435bc1d217eee67c1daf73833258c2035be3df

PR #54, opened 2015-07-24
use -P dll_macro_name on Windows to fix linking problem

dirk-thomas:
Connects to ros2/ros2#26


wjwwood:
+1, with passing Linux job.

dirk-thomas:
Linux job queued: http://ci.ros2.org/job/ros2_batch_ci_linux/83/

PR #55, opened 2015-07-24
Pass timeout to rmw_wait

jacquelinekay:
Connects to ros2/ros2#73

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -812,7 +812,7 @@ rmw_wait(
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
-  bool non_blocking)
+  rmw_time_t * wait_timeout)
 {
   DDS::WaitSet waitset;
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This needs a rebase since the commit list contains other commits.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -878,10 +878,17 @@ rmw_wait(
...
+  } else {
+    non_blocking = wait_timeout->sec == 0 && wait_timeout->nsec == 0;
+    timeout.sec = wait_timeout->sec;
+    timeout.nanosec = wait_timeout->nsec;
+  }
   DDS::ReturnCode_t status = DDS::RETCODE_TIMEOUT;
+
   while (DDS::RETCODE_TIMEOUT == status) {
     status = waitset.wait(active_conditions, timeout);
     if (DDS::RETCODE_TIMEOUT == status) {
```
jacquelinekay: As I was writing a test for this, I realized that the existing logic of this loop is not what I thought it was.

Old behavior:

* Get "blocking" or "non-blocking in rclcpp.
* In rmw implementation:

* if non-blocking, call rmw_wait with a timeout of zero, and don't loop if timed out.
* if blocking: while return code is timeout, wait for 1 second. throw if error, continue if retcode was not timeout. The executor will always block indefinitely inside of rmw_wait.

Here's what I think the new behavior should be:

Get timeout from rclcpp.
* If NULL: block indefinitely in rmw_wait, by passing `DURATION_INFINITE` to `WaitSet::Wait`.
* If positive: pass the timeout value to `WaitSet::wait` and return if timed out.
* If non-blocking: same behavior as before.

Thoughts on this? (@wjwood?)

wjwwood: We discussed this off-line, and it seems that the `DURATION_INFINITE` is the right approach, since no additional checks are being done in the busy wait loop anyways. We're relying on our guard condition to wake the waitset up in an exit condition.

dirk-thomas:
+1

esteve:
+1

wjwwood:
+1

tfoote:
+1 with squash/rebase

PR #57, opened 2015-07-29
fix returning from rmw_wait while any samples have not been taken yet

dirk-thomas:
Same fix for the subscription conditions as ros2/rmw_connext#62


wjwwood:
+1, good catch on the missing check of the return value of the `data_reader`.

esteve:
+1, thanks for working on this!

PR #58, opened 2015-07-31
link rmw_opensplice against pthread

codebot:
On Ubuntu 15.04, I found it necessary to add -lpthread . Not sure why, or where the best place is to add this link, but anyway if it goes here, the examples seem to link and run correctly.


dirk-thomas:
For one branch in the CMake module of OpenSplice we explicitly set that library dependency: https://github.com/ros2/rmw_opensplice/blob/f53c922e2a2de02fe15cfdd843b7c4ad3cb7302f/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake#L85

Therefore I think the patch should be applied somewhere in the alternative branch of the module or in the CMake config file provided by the OpenSplice package (and not in the rmw_opensplice_cpp package). Can you please describe how you installed OpenSplice.

wjwwood:
We talked about this in the meeting today, bump @codebot

PR #59, opened 2015-08-03
create new rmw functions for topic information without implementation

dirk-thomas:
Connects to ros2/ros1_bridge#3

This only adds stubs for the new rmw functions.


wjwwood:
+1 for merge, if a follow on ticket is opened.

tfoote:
ditto @wjwwood

dirk-thomas:
And finally implemented.

PR #62, opened 2015-08-08
modify generated opensplice code to compile with VS 2015

dirk-thomas:
Connects to ros2/ros2#72


wjwwood:
+1

tfoote:
+1 are we tracking submitting this upstream?

dirk-thomas:
I am not planning to since upstream does not support anything newer then Visual Studio 2010. If anybody would like to please go ahead.

PR #64, opened 2015-08-12
Check error codes in attach_condition

jacquelinekay:
Improve the error message for RMW_SET_ERROR_MSG based on different DDS return codes for `Waitset::attach_condition`.

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -195,6 +195,21 @@ struct OpenSpliceStaticServiceInfo
   const service_type_support_callbacks_t * callbacks_;
 };
 
+rmw_ret_t check_attach_condition_error(DDS::ReturnCode_t retcode)
+{
+  if (retcode == DDS::RETCODE_OK) {
+    return RMW_RET_OK;
+  }
+  if (retcode == DDS::RETCODE_OUT_OF_RESOURCES) {
+    RMW_SET_ERROR_MSG("failed to attach condition to waitset: out of resources");
+  } else if (retcode == DDS::RETCODE_BAD_PARAMETER) {
+    RMW_SET_ERROR_MSG("failed to attach condition to waitset: condition pointer was invalid");
+  } else {
+    RMW_SET_ERROR_MSG("failed to attach condition to waitset");
+  }
+  return RMW_RET_ERROR;
+}
+
 const char *
 rmw_get_implementation_identifier()
 {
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1126,7 +1126,16 @@ rmw_wait(
...
       return RMW_RET_ERROR;
     }
-    if (waitset.attach_condition(guard_condition) != DDS::RETCODE_OK) {
+    DDS::ReturnCode_t retcode = waitset.attach_condition(guard_condition);
+    if (retcode == DDS::RETCODE_OUT_OF_RESOURCES) {
+      RMW_SET_ERROR_MSG("failed to attach condition to waitset: out of resources");
+      return RMW_RET_ERROR;
+    }
+    if (retcode == DDS::RETCODE_BAD_PARAMETER) {
+      RMW_SET_ERROR_MSG("failed to attach condition to waitset: invalid condition");
```
wjwwood: There is a more detailed description from of the error from the OpenSplice docs: `the parameter cond is not a valid Condition_ptr`

jacquelinekay: I'll change it to something like "condition pointer was invalid". The parameter `cond` bit is probably too specific for this error message.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1126,7 +1126,16 @@ rmw_wait(
...
+      return RMW_RET_ERROR;
+    }
+    if (retcode != DDS::RETCODE_OK) {
```
dirk-thomas: Since these condition and code blocks are overlapping I would suggest the following structure to shorten the logic branches:

    if (retcode != DDS::RETCODE_OK) {
      if (retcode == DDS::RETCODE_OUT_OF_RESOURCES) {
        RMW_SET_ERROR_MSG(...);
      } else if (retcode == DDS::RETCODE_BAD_PARAMETER) {
        RMW_SET_ERROR_MSG(...);
      } else {
        RMW_SET_ERROR_MSG(...);
      }
      return RMW_RET_ERROR;
    }

tfoote:
+1

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1112,8 +1127,8 @@ rmw_wait(
       RMW_SET_ERROR_MSG("read condition handle is null");
       return RMW_RET_ERROR;
     }
-    if (waitset.attach_condition(read_condition) != DDS::RETCODE_OK) {
-      RMW_SET_ERROR_MSG("failed to attach condition to waitset");
+    DDS::ReturnCode_t retcode = waitset.attach_condition(read_condition);
+    if (check_attach_condition_error(retcode) == RMW_RET_ERROR) {
```
dirk-thomas: We should check for `!= RMW_RET_OK` instead since there are other error values.

Same below.

esteve:
+1

dirk-thomas:
+1

PR #67, opened 2015-08-17
Accept QoS profiles via an XML file

esteve:
Connect to ros2/ros2#86

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -551,6 +551,8 @@ rmw_create_publisher(
     case RMW_QOS_POLICY_KEEP_ALL_HISTORY:
       datawriter_qos.history.kind = DDS::KEEP_ALL_HISTORY_QOS;
       break;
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -202,8 +202,14 @@ rmw_get_implementation_identifier()
 }
 
 rmw_ret_t
-rmw_init()
+rmw_init(const char * qos_xml_filename)
 {
+  std::string qos_xml_filename_str(qos_xml_filename);
+  if (!qos_xml_filename_str.empty()) {
+    RMW_SET_ERROR_MSG("rmw_opensplice does not support passing QoS profiles via XML files");
```
esteve: Unfortunately, the only way of passing an XML file with the QoS profiles to OpenSplice is via the `QosProvider` API, which is only available in the `isocpp` variant. We're currently using `dcps`, which is incompatible with `isocpp`

dirk-thomas:
Closing due to https://github.com/ros2/ros2/issues/86#issuecomment-140549073

PR #72, opened 2015-08-22
fix  clean up participant before deleting.

tfoote:
Fix cleanup on node destruction. 

connects to ros2/system_tests#10

fixes the ticket  ros2/system_tests#29

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -419,7 +419,15 @@ rmw_destroy_node(rmw_node_t * node)
     RMW_SET_ERROR_MSG("participant handle is null");
     return RMW_RET_ERROR;
   }
+
   auto result = RMW_RET_OK;
+  // This unregisters types and destroys topics which were shared between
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -419,6 +419,11 @@ rmw_destroy_node(rmw_node_t * node)
     RMW_SET_ERROR_MSG("participant handle is null");
     return RMW_RET_ERROR;
   }
+
+  // This unregisters types and destroys topics which were shared between
+  // publishers and subscribers and could not be cleaned up in the delete functions.
+  participant->delete_contained_entities();
```
dirk-thomas: Please handle the return value.

tfoote:
updated to catch return code

esteve:
+1

dirk-thomas:
+1

Should this also be done for the rmw_connext implementations? Or is that done automatically for Connext but now for OpenSplice?

tfoote:
The original version of this PR without the error handling is the state in rmw_connext.

PR #75, opened 2015-08-24
implement the rmw_gid_t related functions

wjwwood:
Connects to ros2/ros2#91

See: https://github.com/ros2/rmw/pull/35

This still needs to be tested. It seems to work, but I'm not 100% certain that the comparison is sufficient. I'm looking for a better wait to compare the OpenSplice GID's.

### DIFF
--- a/rmw_opensplice_cpp/src/functions.cpp
+++ b/rmw_opensplice_cpp/src/functions.cpp
@@ -23,6 +23,7 @@
 
 #include <ccpp_dds_dcps.h>
 #include <dds_dcps.h>
+#include <u_instanceHandle.h>
 
 #include <rmw/allocators.h>
 #include <rmw/error_handling.h>
@@ -202,12 +203,18 @@ struct OpenSpliceStaticNodeInfo
   CustomSubscriberListener * subscriber_listener;
 };
 
+typedef struct OpenSplicePublisherGID
+{
+  DDS::InstanceHandle_t publication_handle;
+} OpenSplicePublisherGID;
+
 struct OpenSpliceStaticPublisherInfo
 {
   DDS::Topic * dds_topic;
   DDS::Publisher * dds_publisher;
   DDS::DataWriter * topic_writer;
   const message_type_support_callbacks_t * callbacks;
+  rmw_gid_t publisher_gid;
 };
 
 struct OpenSpliceStaticSubscriberInfo
@@ -612,6 +619,18 @@ rmw_create_publisher(
   publisher_info->dds_publisher = dds_publisher;
   publisher_info->topic_writer = topic_writer;
   publisher_info->callbacks = callbacks;
+  static_assert(
+    sizeof(OpenSplicePublisherGID) <= RMW_GID_STORAGE_SIZE,
+    "RMW_GID_STORAGE_SIZE insufficient to store the rmw_opensplice_cpp GID implemenation."
+  );
+  // Zero the data memory.
+  memset(publisher_info->publisher_gid.data, 0, RMW_GID_STORAGE_SIZE);
+  {
+    auto publisher_gid =
+      reinterpret_cast<OpenSplicePublisherGID *>(publisher_info->publisher_gid.data);
+    publisher_gid->publication_handle = topic_writer->get_instance_handle();
+  }
+  publisher_info->publisher_gid.implementation_identifier = opensplice_cpp_identifier;
 
   publisher->implementation_identifier = opensplice_cpp_identifier;
   publisher->data = publisher_info;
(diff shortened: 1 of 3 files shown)


rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1034,53 +1039,80 @@ rmw_destroy_subscription(rmw_node_t * node, rmw_subscription_t * subscription)
 rmw_ret_t
 rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * taken)
 {
+  auto publisher_gid = rmw_take_with_gid(subscription, ros_message, taken);
+  bool result = (publisher_gid != nullptr);
+  rmw_destroy_gid(publisher_gid);
```
dirk-thomas: I don't like the idea that every `rmw_take` creates a gid and destroys it immediately again.

wjwwood: I agree, but this was the shortest path. I'll refactor it into three functions, one that does the work but also returns the information needed to the create the publisher_gid, `take` which is normal, but doesn't create a gid, and take with gid which uses the extra info to make a gid too.

wjwwood: Ok, have a look at eecd08c.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1083,6 +1090,40 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * tak
...
+  rmw_gid_t * sender_gid = static_cast<rmw_gid_t *>(rmw_allocate(sizeof(rmw_gid_t)));
+  if (!sender_gid) {
+    RMW_SET_ERROR_MSG("failed to allocate memory");
+    return nullptr;
+  }
+  OpenSplicePublisherGID * detail =
+    static_cast<OpenSplicePublisherGID *>(rmw_allocate(sizeof(OpenSplicePublisherGID)));
+  if (!detail) {
+    RMW_SET_ERROR_MSG("failed to allocate memory");
+    return nullptr;
```
dirk-thomas: Needs to free `sender_gid`.

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1083,6 +1090,40 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * tak
...
 }
 
+rmw_ret_t
+rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * taken)
+{
+  return _take(subscription, ros_message, taken, nullptr);
+}
+
+rmw_gid_t *
+rmw_take_with_gid(const rmw_subscription_t * subscription, void * ros_message, bool * taken)
```
dirk-thomas: How does this function signal the specific error code? Shouldn't it return `rmw_ret_t` and have an additional argument `rmw_gid_t * &`?

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -2094,4 +2135,92 @@ rmw_count_subscribers(
...
+  detail->publisher_system_id = topic_writer_gid.systemId;
+  gid->implementation_identifier = opensplice_cpp_identifier;
+  gid->data = detail;
+  return gid;
+}
+
+void
+rmw_destroy_gid(rmw_gid_t * gid)
+{
+  if (gid) {
```
dirk-thomas: This must check the `implementation_identifier`.

dirk-thomas:
The gid is one additional meta information we add in this PR. I expect that we might want to add another piece of information in the future. May be we should encapsulate the gid in a struct similar to the DDS `SampleInfo`.

wjwwood:
I agree. I was thinking of making `rmw_publisher_info_t` and `rmw_message_info_t` or `rmw_message_instance_info_t` classes, but I think they would both contain a `rmw_gid_t`, so I avoided that for now as optional.

rmw_opensplice_cpp/src/functions.cpp
wjwwood: It returns a nullptr to indicate an error and sets the error string. This is a pretty common thing across all of our functions. If we want to return an error specific error code we need to refactor many of our rmw functions. I don't want to do that here.

rmw_opensplice_cpp/src/functions.cpp
wjwwood: Yup, fixed in 9e4bb2f.

rmw_opensplice_cpp/src/functions.cpp
wjwwood: Fixed in b9db525 and ros2/rmw@be65bf6

jacquelinekay:
+1

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -1083,6 +1106,39 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * tak
...
+    return RMW_RET_ERROR;
+  }
+  DDS::InstanceHandle_t sending_publication_handle;
+  auto ret = _take(subscription, ros_message, taken, &sending_publication_handle);
+  if (ret != RMW_RET_OK) {
+    // Error string is already set.
+    return RMW_RET_ERROR;
+  }
+
+  rmw_gid_t * sender_gid = const_cast<rmw_gid_t *>(&message_info->publisher_gid);
```
dirk-thomas: Neither `message_info` nor `publisher_gid` seems to be const. Why is the const_cast necessary here?

dirk-thomas:
LGTM

tfoote:
+1

rmw_opensplice_cpp/src/functions.cpp
wjwwood: Const was dropped in a refactoring. 615b37e

wjwwood:
Squashed.

PR #83, opened 2015-09-23
Bypass OpenSplice query encoding

esteve:
This uses standard C++ string concatenation instead of leaving it up to OpenSplice to encode the query. OpenSplice has issues with encoding arguments that generate queries that are too long.

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
+++ b/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
@@ -54,10 +54,17 @@ public:
     writer_guid_.first = uniform_dist(e1);
     writer_guid_.second = uniform_dist(e1);
 
+    // NOTE(esteve): OpenSplice's query encoding on Windows seems to have issues with
+    // arguments that produce queries that are too long
+    // https://github.com/PrismTech/opensplice/issues/21
+    std::stringstream ss;
+    ss << "client_guid_0_ = " << writer_guid_.first << " AND client_guid_1_ = " <<
+      writer_guid_.second;
+
+    std::string query(ss.str());
+
     DDS::StringSeq args;
-    args.length(2);
-    args[0] = DDS::string_dup(std::to_string(writer_guid_.first).c_str());
-    args[1] = DDS::string_dup(std::to_string(writer_guid_.second).c_str());
+    args.length(0);
 
     DDS::ReturnCode_t status;
     DDS::TopicQos default_topic_qos;
@@ -133,7 +140,7 @@ public:
     // Let OpenSplice do any needed encoding
     content_filtered_response_topic_ = participant_->create_contentfilteredtopic(
       service_name_.c_str(), response_topic_,
-      "client_guid_0_ = %0 AND client_guid_1_ = %1",
+      query.c_str(),
       args);
     if (!content_filtered_response_topic_) {
       estr = "DomainParticipant::create_contentfilteredtopic: failed";


wjwwood:
Wow that stinks. Is there an upstream issue? Is there an error or does it just silently fail?

esteve:
I'll file a ticket upstream.

Some times (but unfortunately not always), there's a `Set query parameter[0] failed` error message, however the `ContentFilteredTopic` object is still created, otherwise this would have failed:

https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp#L138

but since I changed this code to use string concatenation, I haven't seen the aforementioned error and the tests pass.

dirk-thomas:
Can you please run a CI job for this and point to the exact test(s) which work after this patch.

esteve:
http://ci.ros2.org/job/ros2_batch_ci_linux/380/testReport/%28root%29/test_services_cpp__rmw_opensplice_cpp/
http://ci.ros2.org/job/ros2_batch_ci_osx/281/testReport/%28root%29/test_services_cpp__rmw_opensplice_cpp/
http://ci.ros2.org/job/ros2_batch_ci_windows/426/testReport/%28root%29/test_services_cpp__rmw_opensplice_cpp/

I re-pushed this branch as `test-services` so that the farm would group them together. The failing tests for connext seem to be caused by something else, I'm working on it.

esteve:
Filed ticket upstream:

https://github.com/PrismTech/opensplice/issues/21

esteve:
This should be ready to be merged. Anything that needs to be changed?

dirk-thomas:
Based on this PR I looked a bit deeper into the OpenSplice code. As far as I can see it relies internally in a lot of cases of these expressions with placeholders. Also the code handling those (parsing, replacement, etc.) has no platform dependent code.

Therefore I don't understand why this should be broken in OpenSplice itself and only on Windows but not on Linux / OS X. Based on that I have created an alternative branch (https://github.com/ros2/rmw_opensplice/commit/6b09dabc14097e7748a9086d9f45c0250f2e3eb1). My assumption for the patch was:
* OpenSplice uses the feature internally so it might not be broken
* we fill the parameters early, do other stuff, and then use them
* maybe some memory corruption is happening in between??

The CI job for that also passed the specific test: @esteve mentioned above: http://ci.ros2.org/job/ros2_batch_ci_windows/430/testReport/%28root%29/test_services_cpp__rmw_opensplice_cpp/

I am not sure if the test is the "right" check for the problem, if it is always failing before, and if if passes always with this change. But based on that I would not merge this PR.

dirk-thomas:
Another option could be that whatever we do here is not actually effecting that test...

esteve:
I was surprised as well because I didn't see anything platform specific. However the code on the GitHub does not seem to reflect the behavior of the binaries in some cases, mainly query parsing (e.g. I couldn't pass integers in hex format, but the code seems to support it):

https://github.com/PrismTech/opensplice/blob/master/src/api/dcps/gapi/code/gapi_common.c#L258

And we had a case in which our code crashed on Windows, but it didn't on Linux (at least not the on the farm). There was nothing platform specific, but due to different timing, references to shared pointers made our tests crash on Windows.

In any case, if there's a cleaner way to fix this, I'm more than happy to close this PR. And if someone else wants to take over the Windows stuff, I'd be even happier :smile:

esteve:
I've updated the description of the PR to denote that it may not be a bug in OpenSplice's Windows binary, but a bug in our code.

dirk-thomas:
I still don't think this should be merged since it clearly does not address the problem but is a workaround (which is yet clearly why it actually works).

esteve:
@dirk-thomas I didn't argue to merge this now, just clarified the PR description and give more insight about what's the context.

wjwwood:
Todo:

- Run @dirk-thomas's patch multiple times to see if its fix is flaky.
- If it passes consistently then:
 - Retest the without patch to assert that it isn't also flaky.

Collect info and talk about it next time and look at our code and opensplice code in between.

wjwwood:
Another option, try building opensplice from source on windows to see if that affects the behavior of the issue.

esteve:
After applying @dirk-thomas 's patch, the services test exhibits the same behavior. In both cases, when sufficiently long GUIDs, OpenSplice creates a `ContentFilteredTopic`, but with the query broken. So I imagine that the test passed for @dirk-thomas because the requester happened to have generated a short GUID. As an example, with `writer_guid_.first = 10604530178203912464` and `writer_guid_.second = 14645646989687087260` and @dirk-thomas's patch, this is the output from OpenSplice:

```
================================================================================
========
Report      : ERROR
Date        : Thu Oct 08 15:47:43 Pacific Daylight Time 2015
Description : Set query parameter[0] failed
Node        : win81-virtual
Process     : test_services_client_cpp__rmw_opensplice_cpp.exe <3948>
Thread      : 268
Internals   : V6.4.140407OSS/3b83e82/3b83e82/Database predicate/c_querybase.c/16
22/0/1444344463.994794036
================================================================================
========
Report      : ERROR
Date        : Thu Oct 08 15:47:43 Pacific Daylight Time 2015
Description : Set query parameter[1] failed
Node        : win81-virtual
Process     : test_services_client_cpp__rmw_opensplice_cpp.exe <3948>
Thread      : 268
Internals   : V6.4.140407OSS/3b83e82/3b83e82/Database predicate/c_querybase.c/16
22/0/1444344463.995950650
```

however, with my patch and the same GUIDs, the test does pass, so I propose merging this PR.

esteve:
@dirk-thomas @jacquelinekay @wjwwood ping

wjwwood:
+1, but please link to the upstream issue (https://github.com/PrismTech/opensplice/issues/21) from your comment in the code. That way later someone can follow up on the upstream issue and potentially clear the todo, going back to using their mechanism.

dirk-thomas:
+/-0 from me since I still don't think this is the way to fix the problem. Having a minimal example demonstrating the bug using the OpenSplice API directly (without any ROS 2 stuff) would be something that convinces me.

We currently also lack the necessary "infrastructure" to run these tests efficiently. If we rely on random numbers to be in a certain range we need to run the tests numerous times to get a reasonable confidence that a certain change improves things.

esteve:
@dirk-thomas I'm going to merge this after applying @wjwwood comment anyway. This solves our problem because bypasses OpenSplice's query encoder for queries with long arguments. Does this fix the bug in OpenSplice? No, it's not our responsibility, the best we can do is report it. And it's completely reasonable to think that there's a bug in how arguments are encoded in OpenSplice, since the approach in the PR takes a different path (generating the query ourselves) and it works.

PR #84, opened 2015-09-23
Added logic to set durability QoS settings

esteve:
This PR adds the logic needed to apply the durability policy to data readers and writers.

Connects to ros2/rmw#41


dirk-thomas:
+1

PR #85, opened 2015-09-25
increase fragment size

dirk-thomas:
Prevent fragmentation as much as possible to work with FastRTPS in more cases for now.

Makes the last interop test pass: http://ci.ros2.org/job/ros2_batch_ci_linux/395/

Connect to ros2/ros2#124


wjwwood:
Seems fine to me, but I'm not sure what happens on machines with smaller UDP packet size settings.

dirk-thomas:
The OS will start fragmenting the packets. That implies that DDS has to resend the full DDS-level fragment since it is not aware of the subfragments introduced by the OS.

PR #86, opened 2015-10-15
update code to pass ament_cpplint

dirk-thomas:


wjwwood:
+1

esteve:
+1

rmw_opensplice_cpp/src/functions.cpp
```diff
@@ -20,27 +20,26 @@
...
-#include <rmw/rmw.h>
-#include <rosidl_generator_c/message_type_support.h>
-#include <rosidl_generator_c/service_type_support.h>
-#include <rosidl_typesupport_opensplice_cpp/identifier.hpp>
-#include <rosidl_typesupport_opensplice_cpp/impl/error_checking.hpp>
-#include <rosidl_typesupport_opensplice_cpp/message_type_support.h>
-#include <rosidl_typesupport_opensplice_cpp/service_type_support.h>
+#include <ccpp_dds_dcps.h>
+#include <dds_dcps.h>
+#include <u_instanceHandle.h>
```
dirk-thomas: About these lines cpplint complains (to be in wrong order). And using quotes instead doesn't make it much better - then cpplint complains about not using a folder name before the filename.

wjwwood: That's less than desirable. I think they might be ok in the other ones because we have some preprocessor stuff surrounding them. Maybe it just gives a pass to anything which is surrounded by preprocessor directives.

PR #88, opened 2015-10-27
Use four 32-bit signed parts to represent the client id

esteve:
This PR replaces unsigned 64-bit integers with signed 32-bit integers to ensure portability.

Connects to ros2/rmw_connext#31


dirk-thomas:
Can you explain why `int32` is more portable then `uint64`?

esteve:
Connext's query parser doesn't seem to accept `uint64` values, but `int32` worked all the time in my tests.

mikaelarguedas:
deleting this based on https://github.com/ros2/rosidl_dds/pull/28#issuecomment-321093757

PR #90, opened 2015-11-03
rmw include directories were missing from typesupport library

wjwwood:
While testing the `rclcpp_library` branch on Windows (using `--isolated`) I found that `test_communications` failed to include `rmw/rmw.h` when building it's type support library for the messages it generates. I can neither explain why this only happens on Windows nor why it doesn't affect other message packages.

Connects to ros2/rclcpp#48


esteve:
+1

PR #92, opened 2015-11-04
split functions.cpp into separate files

dirk-thomas:
The code has only been moved but not modified (except visibility).

http://ci.ros2.org/job/ros2_batch_ci_linux/548/
http://ci.ros2.org/job/ros2_batch_ci_osx/416/
http://ci.ros2.org/job/ros2_batch_ci_windows/565/

PR #93, opened 2015-11-06
split functions.cpp into separate files (with history)

dirk-thomas:
This replaces #92 and has identical changes exept with a different commit history to preserve the history. It allows to blame changes from before the move, e.g. https://github.com/ros2/rmw_opensplice/blame/ff9fd166df2b1f9ec82f23f39d3da4c9ef261002/rmw_opensplice_cpp/src/rmw_client.cpp


esteve:
+1

Issue #47, opened 2015-07-15
Create teardown functions for requester and replier


wjwwood:
This is something that is left over from #45, and I left a TODO's for:

- https://github.com/ros2/rmw_opensplice/blob/75fd4523b8237f98964815abe8c84fa3d70a7ae3/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template#L537
- https://github.com/ros2/rmw_opensplice/blob/75fd4523b8237f98964815abe8c84fa3d70a7ae3/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template#L562

Basically we need to have a matching `deinit` method on those classes which properly undoes what is created/initialized in the `init` methods. Also like the `init` method it should be `noexcept` and It should check all of the error cases and return an error string in the event of failure.

PR #95, opened 2015-11-12
Teardown functions

jacquelinekay:
Fixes ros2/rmw_opensplice#47

This also fixes a bug with our implementation of Opensplice requesters that I discovered while using multiple clients on the same topic by giving the ContentFilteredTopic a unique name based on the writer GUID of the requester.

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
+++ b/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
@@ -75,6 +75,8 @@ public:
     std::string request_topic_name = service_name_ + "_Request";
     std::string response_type_name = service_type_name_ + "_Response_";
     std::string response_topic_name = service_name_ + "_Response";
+    std::string content_filtered_topic_name = service_name_ +
+      std::to_string(writer_guid_.first) + "_" + std::to_string(writer_guid_.second);
     const char * estr = nullptr;
     request_topic_ = nullptr;
     request_publisher_ = nullptr;
@@ -137,10 +139,8 @@ public:
       estr = "DomainParticipant::create_topic: failed for response";
       goto fail;
     }
-
-    // Let OpenSplice do any needed encoding
     content_filtered_response_topic_ = participant_->create_contentfilteredtopic(
-      service_name_.c_str(), response_topic_,
+      content_filtered_topic_name.c_str(), response_topic_,
       query.c_str(),
       args);
     if (!content_filtered_response_topic_) {
@@ -157,12 +157,6 @@ public:
     }
     return nullptr;
 fail:
-    if (content_filtered_response_topic_) {
-      status = participant_->delete_contentfilteredtopic(content_filtered_response_topic_);
-      if (nullptr != impl::check_delete_contentfilteredtopic(status)) {
-        fprintf(stderr, "%s\n", impl::check_delete_contentfilteredtopic(status));
-      }
-    }
     if (response_datareader_) {
       // Assumption: subscriber is not null at this point.
       status = response_subscriber_->delete_datareader(response_datareader_);
@@ -174,12 +168,6 @@ fail:
       status = participant_->delete_subscriber(response_subscriber_);
       if (nullptr != impl::check_delete_subscriber(status)) {
         fprintf(stderr, "%s\n", impl::check_delete_subscriber(status));
-      }
-    }
-    if (response_topic_) {
-      status = participant_->delete_topic(response_topic_);
-      if (nullptr != impl::check_delete_topic(status)) {
-        fprintf(stderr, "%s\n", impl::check_delete_topic(status));
       }
     }
     if (request_datawriter_) {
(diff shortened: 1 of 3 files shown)


rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
```diff
@@ -164,6 +164,49 @@ class Responder
...
+      if (nullptr != impl::check_delete_subscriber(status)) {
+        fprintf(stderr, "%s\n", impl::check_delete_subscriber(status));
+      }
+    }
+    if (request_topic_) {
+      status = participant_->delete_topic(request_topic_);
+      if (nullptr != impl::check_delete_topic(status)) {
+        fprintf(stderr, "%s\n", impl::check_delete_topic(status));
+      }
+    }
```
wjwwood: If this is identical to the fail section of `init`, then I'd suggest making it a function and avoiding duplicate code.

However, it seems like the `deinit` function could return a string if there is there is a problem destructing things, similar to this:

https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/src/rmw_subscription.cpp#L242-L280

Basically you try each step (even if there are failures along the way) and if something fails you set the error string. But what's not captured clearly in the above link is that if the error string is already set (more than one thing failed) then the previous error string that's getting replaced is printed to stderr. So the `deinit` function could behave like this:

- Everything worked, return nullptr
- One thing failed, try all steps (don't return on failure) and return the error string for the thing that failed
- Multiple things failed, same as the previous point, but each time a new thing fails, print the error to stderr, and return the last error string through the normal channel

But I can tell you from experience that this is a rabbit hole of grace degradation that takes a lot of effort, so I think at some point just printing these problems to stderr is fine. The only exception to this is that the `deinit` function should probably cause the calling code to fail it it failed. This is the case for the `init` function: https://github.com/ros2/rmw_opensplice/pull/95/files#diff-cf8d0757343d0e3c757d5509dc29f18dR381

jacquelinekay: OK, since `deinit` gets called from `destroy_requester__<srv_name>` and `destroy_responder__<srv_name>` I will change it to return an error string and cause those functions to fail if the string is not null.

also, this is a total bikeshed note, but I'm going to change the name from `deinit` to `teardown` or something else that's more specific.

wjwwood: Sounds good to me.

jacquelinekay: I see what you meant about a rabbit hole.

jacquelinekay:
Jenkins (Windows is currently offline):

http://ci.ros2.org/job/ros2_batch_ci_linux/589/
http://ci.ros2.org/job/ros2_batch_ci_osx/475/

Linux shows an unrelated cpplint error in rosidl_generator_c.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
```diff
@@ -137,10 +139,8 @@ class Requester
       estr = "DomainParticipant::create_topic: failed for response";
       goto fail;
     }
-
-    // Let OpenSplice do any needed encoding
     content_filtered_response_topic_ = participant_->create_contentfilteredtopic(
-      service_name_.c_str(), response_topic_,
+      content_filtered_topic_name.c_str(), response_topic_,
```
esteve: This shouldn't be necessary. The purpose of using a `ContentFilteredTopic` is such that we reuse the same topic name, but let the filtering happen on the content (hence the query). DDS would inspect the fields of a message and if they match the query, the message will be routed to the appropriate client.

jacquelinekay: When I wrote a test case with multiple clients listening on the same topic, the case failed for Opensplice and passed both Connext implementations. After some debugging it was apparent that I couldn't have multiple ContentFilteredTopic objects with the same name (which was happening because I had multiple clients passing the same service name to the ContentFilteredTopic objects). So I changed the topic name to `ServiceName_writer_guid.first_writer_guid.second` (as you can see above) so that each client would have a unique ContentFilteredTopic name.

esteve: Connext uses a `ContentFilteredTopic` with the same topic name and filtered on the content too, otherwise the query wouldn't be necessary and we could just use a regular `Topic` object. I haven't had a deeper look at it, but it appears the bug must be somewhere else.

esteve: I just checked with `rtiddspy` and you are right, Connext creates a `ContentFilteredTopic` whose topic name is different per requester.

esteve:
+1

PR #96, opened 2015-11-24
Use global WaitSet to be reused in rmw_wait

jacquelinekay:
In progress. I'd appreciate some preliminary feedback on this approach.

TODO:

Global ConditionSeq to be used rmw_wait (with option to pre-allocate)
Matching updates to Connext
Should the global shared variables be initialized/de-initialized in rmw_init/rmw_shutdown?

### DIFF
--- a/rmw_opensplice_cpp/CMakeLists.txt
+++ b/rmw_opensplice_cpp/CMakeLists.txt
@@ -47,6 +47,7 @@ add_library(rmw_opensplice_cpp SHARED
   src/rmw_trigger_guard_condition.cpp
   src/rmw_wait.cpp
   src/types.cpp
+  src/waitset.cpp
 )
 ament_target_dependencies(rmw_opensplice_cpp
   "rmw"
--- a/rmw_opensplice_cpp/src/rmw_wait.cpp
+++ b/rmw_opensplice_cpp/src/rmw_wait.cpp
@@ -21,6 +21,8 @@
 
 #include "types.hpp"
 
+#include "waitset.hpp"
+
 // The extern "C" here enforces that overloading is not used.
 extern "C"
 {
@@ -47,7 +49,7 @@ rmw_wait(
   rmw_clients_t * clients,
   rmw_time_t * wait_timeout)
 {
-  DDS::WaitSet waitset;
+  //DDS::WaitSet waitset;
 
   // add a condition for each subscriber
   for (size_t i = 0; i < subscriptions->subscriber_count; ++i) {
@@ -63,7 +65,7 @@ rmw_wait(
       return RMW_RET_ERROR;
     }
     rmw_ret_t status = check_attach_condition_error(
-      waitset.attach_condition(read_condition));
+      opensplice_cpp_waitset.attach_condition(read_condition));
     if (status != RMW_RET_OK) {
       return status;
     }
@@ -78,7 +80,7 @@ rmw_wait(
       return RMW_RET_ERROR;
     }
     rmw_ret_t status = check_attach_condition_error(
-      waitset.attach_condition(guard_condition));
+      opensplice_cpp_waitset.attach_condition(guard_condition));
     if (status != RMW_RET_OK) {
       return status;
     }
@@ -98,7 +100,7 @@ rmw_wait(
       return RMW_RET_ERROR;
     }
     rmw_ret_t status = check_attach_condition_error(
-      waitset.attach_condition(read_condition));
+      opensplice_cpp_waitset.attach_condition(read_condition));
     if (status != RMW_RET_OK) {
       return status;
     }
@@ -118,7 +120,7 @@ rmw_wait(
       return RMW_RET_ERROR;
     }
     rmw_ret_t status = check_attach_condition_error(
-      waitset.attach_condition(read_condition));
+      opensplice_cpp_waitset.attach_condition(read_condition));
     if (status != RMW_RET_OK) {
       return status;
     }
@@ -133,7 +135,7 @@ rmw_wait(
     timeout.sec = static_cast<DDS::Long>(wait_timeout->sec);
     timeout.nanosec = static_cast<DDS::Long>(wait_timeout->nsec);
   }
-  DDS::ReturnCode_t status = waitset.wait(active_conditions, timeout);
+  DDS::ReturnCode_t status = opensplice_cpp_waitset.wait(active_conditions, timeout);
 
   if (status == DDS::RETCODE_TIMEOUT) {
     return RMW_RET_TIMEOUT;
@@ -153,6 +155,7 @@ rmw_wait(
       return RMW_RET_ERROR;
     }
     DDS::ReadCondition * read_condition = subscriber_info->read_condition;
+    opensplice_cpp_waitset.detach_condition(read_condition);
     if (!read_condition) {
       RMW_SET_ERROR_MSG("read condition handle is null");
       return RMW_RET_ERROR;
@@ -173,6 +176,7 @@ rmw_wait(
       return RMW_RET_ERROR;
     }
 
+    opensplice_cpp_waitset.detach_condition(guard_condition);
     if (!guard_condition->get_trigger_value()) {
       // if the guard condition was not triggered
       // reset the guard condition handle
@@ -199,6 +203,7 @@ rmw_wait(
       RMW_SET_ERROR_MSG("read condition handle is null");
       return RMW_RET_ERROR;
     }
+    opensplice_cpp_waitset.detach_condition(read_condition);
 
     // search for service condition in active set
     uint32_t j = 0;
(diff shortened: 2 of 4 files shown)


dirk-thomas:
Please do not use a global variable. Instead rmw_init() needs to return a handle where each rmw impl. can embed custom information. This handle must then also be passed to rmw_wait().

jacquelinekay:
alright, before I implement that for Opensplice and Connext, here's an outline of the approach:

```
typedef struct rmw_custom_handle {
  void * data;
} rmw_custom_handle_t;

rmw_custom_handle_t * rmw_init();
rmw_ret_t rmw_shutdown(rmw_custom_handle_t * handle);
rmw_ret_t rmw_wait( /* current arguments... */, rmw_custom_handle_t * handle);
```

In `rmw_opensplice_cpp`:

```
struct OpenSpliceCustomHandle {
  DDS::WaitSet waitset;
  // other custom global state
}

rmw_custom_handle_t * rmw_init() {
  OpenSpliceCustomHandle * handle = rmw_allocate(sizeof(OpenSpliceCustomHandle));
  // other initialization on the handle, return null if error
  rmw_custom_handle_t * rmw_handle = rmw_allocate(sizeof(rmw_custom_handle_t));
  return static_cast<void*>(rmw_handle);
}
// similarly, deallocate in rmw_shutdown

rmw_ret_t rmw_wait( /* current arguments... */, rmw_custom_handle_t * handle) {
  OpenSpliceCustomHandle * handle = static_cast<OpenSpliceCustomHandle*>(handle->data);
  DDS::WaitSet * waitset = handle->waitset;
  // ...
}
```

So, rclcpp and other rcl implementations have to store the handle somewhere (probably in Executor or context).

wjwwood:
I don't think having `rmw_init` return something is any different from using a global since you can only call `rmw_init` once.

If we want to be able to call `rmw_wait` on different wait sets simultaneously (in different threads) then I'd recommend having a pair of functions like `rmw_init_wait_storage` and `rmw_fini_wait_storage` which operate on a struct which has an implementation identifier pointer and a `void *` (like all our other structs in `rmw`) and then modify `rmw_wait` take a pointer to that struct. Then the implementations can store what ever they want in that struct and the executor can hold a handle to that which it creates on constructions and finalizes on destruction.

If we want to say that `rmw_wait` is not thread safe and therefore should only ever be called once per process I'd say use a global and an atomic bool for error checking (ensuring `rmw_wait` wasn't called concurrently) rather than further complicating the `rmw` API with another type to be tracked.

Our decision on this matters to me since it is reflected in the `rcl` API.

dirk-thomas:
@wjwwood Why can `rmw_init` only be called once? It should be possible after a corresponding `rmw_fini()` call to start over again, no?

Also depending on the rmw impl. there is no reason to prevent someone to juggle to rmw instances at the same time if their state is fully encapsulated in these handles (rather then using global variables).

@jacquelinekay As William mentioned the struct should also contain an identifier (like the other struct) to ensure the the same rmw impl. is handling it. Eventually the handle should be passed to many if not all functions.

To avoid requiring to pass multiple "handles" to some functions more specific structs like the node handle could reference the generic one. Then only the most specific handle need to be passed.

gbiggs:
I'd like to echo @dirk-thomas 's comment about the multiple rmw instances use case. While it seems unlikely to me that someone would want to use two DDS implementations at the same time, the idea that something other than DDS can provide an rmw implementation has been promoted quite a bit. If I have a pure TCP implementation, a DDS implementation, and (the stuff of nightmares) a CORBA implementation, then I should be able to use them all from within a single node if my system designer or my deployer says that is what is needed.

Also, perhaps, two instances of a statically (compile-time) configurable DDS implementation? I'm not sure such a thing exists, but if it does then the deployer should have the option of using both simultaneously.

If the above use cases can be supported using global variables then great, but if not then I back the choice of encapsulating state in handles.

wjwwood:
`rmw_init` can only be called once at a time (it can be called again after `rmw_shutdown`). It was designed that way because it does not return any state. In my version of `rcl` the `rcl_init` function can only be called once and it is explicitly documented as such. `rmw_init` and `rcl_init` do not have to be symmetric, but as it is currently designed `rmw_init` shouldn't be called more than once per process (except calling it again after shutdown).

I don't see a compelling reason to redesign the `rmw` API such that you can call `rmw_init` multiple times within the same process (without calling `rmw_shutdown` each time), in order to juggle more than one rmw instance at a time. That adds a lot of complexity for very little value in my opinion and currently none of our implementations could handle that, so they'ed all need to be updated to support it.

@gbiggs To be clear @dirk-thomas is not talking about having different rmw implementations being run simultaneously, just multiple instances of the same rmw implementation.

I think the more immediate question is whether to allow `rmw_wait` to be called concurrently on (non-overlapping?) wait sets. The question about having multiple instances of the rmw at a time is something that unrelated to what @jacquelinekay is doing and requires not only API changes but also implementation changes to support it.

gbiggs:
@wjwwood Thanks for the clarification. Your argument makes more sense now. I think that being able to have multiple instances of the same rmw implementation *could* be useful in some corner cases but if it requires both API and implementation changes then it doesn't seem worth the effort.

jacquelinekay:
I agree that it adds unnecessary complexity to support immediately subsequent calls to rmw_init (perhaps from different threads) that instantiate different instances of rmw.

On a different note, I do think there is a compelling argument to have access to the rmw state that is implicitly instantiated by the library (the state that Dirk proposed to encapsulate in a handle) or at least pass argument to it on initialization. For example, I'd like the user to have the ability to preallocate the ConditionSeq object in rmw_wait with the maximum number of conditions that rmw could respond to. I think this quantity is general enough to expose to the rmw interface. The alternative would be to hardcode an arbitrary value, which I find very unappealing.

wjwwood:
I'd be fine having `rmw_init` return a "rmw context" which could also hold instance specific storage to help avoid globals as a sort of future proofing, but only if we document that `rmw_init` cannot be used as described above for now. Then later, once we have a compelling reason to support that, we can make sure all the implementations support it and then change the documentation, but hopefully not the interface. This object would need to be passed to the node and would need to be passed to shutdown and rmw_wait.

wjwwood:
> On a different note, I do think there is a compelling argument to have access to the rmw state that is implicitly instantiated by the library (the state that Dirk proposed to encapsulate in a handle) or at least pass argument to it on initialization. For example, I'd like the user to have the ability to preallocate the ConditionSeq object in rmw_wait with the maximum number of conditions that rmw could respond to. I think this quantity is general enough to expose to the rmw interface. The alternative would be to hardcode an arbitrary value, which I find very unappealing.

Exposing the state in a handle is fine, but we have to be careful to not expose the DDS symbols across the `rmw` boundary which I think would be required to do what you've just described. Without the DDS symbols you'd have to do something like what I did with the GUID type and just say well in DDS it's this size so it will be this size in our struct so our struct can store the DDS object, see: https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h#L34-L36

Then any other implementations would have to work within that constraint.

jacquelinekay:
Ok, I'm going to save the work we've discussed on `rmw_context` for the next pull request unless it turns out to be necessary for adding a global waitset.

We should address the more immediate question that @wjwwood brought up in his previous post: if multiple threads in the same process can enter `rmw_wait` at the same time.

In the current implementation, if two threads try to wait on the same entity (subscription/client/service) at the same time, I believe they will both try to execute the same entity twice, which is a bug (though this is not possible in rclcpp because we don't allow multiple threads to spin at once, and rmw_wait is only called in spin* functions).

But if there are two nodes executing in the same process, perhaps they should be able to enter rmw_wait at the same time, since they do not share instances of subscriptions/clients/services. This would be a compelling reason to have two Executors, to simplify the logic for separating the execution and execution state of the two nodes, while still having one intra-process manager to handle the intra-process messaging between the nodes. In this case, I think each Executor would need its own rmw "context" or "handle".

The question that remains for me is if this case is compelling enough to add more complexity to rmw.

dirk-thomas:
In the current state of the PR the error handling is completely broken since the  now global WaitSet keeps its state. For repeated invocations of rmw_wait it is not guaranteed that the WaitSet has no attached conditions before adding new ones.

jacquelinekay:
I detach the read_conditions and guard_conditions after calling WaitSet::wait in this PR:

https://github.com/ros2/rmw_opensplice/pull/96/files#diff-02d62dc3a10082a1c0f51479c76af097R158

If you think this is broken, can you suggest an alternative way of detaching all conditions from the WaitSet?

dirk-thomas:
Any return statement in this function will prevent the WaitSet to be set back into a state where it doesn't have any conditions attached.

gbiggs:
The use case of having multiple executors in a single process is fairly important for us. It provides more flexibility to node management approaches than each process being restricted to a single executor would.

How much complexity would it add to rmw? As far as I know, DDS itself shouldn't provide any roadblocks.

tfoote:
Investigate whether DDS's wait call can be invoked from two separate threads and under what conditions (e.g., what elements can or cannot be shared across the two wait sets).

Assuming that thread-safe waiting is fundamentally possible, then add APIs to rmw to create and destroy wait sets, passing back and forth handles to them, to be stored, for example, in an executor.

jacquelinekay:
Going to redo this in a forthcoming PR

PR #97, opened 2015-12-02
Make WaitSet static in rmw_wait

gerkey:
This change is intended to prevent a race that can occur between (the exit of) `rmw_wait()` here and the invocation of `rmw_trigger_guard_condition` at https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/executor.cpp#L179.  The theory is that triggering the guard condition causes it to iterate over its internal list (in our case, of length 1) of wait sets to which it is attached, and that this can happen at exactly the time that one of those wait sets is being destroyed in another thread.

To avoid that race, we never allow the wait set to be destroyed, and instead do our own cleanup (detaching conditions) of its contents on exit from `rmw_wait`.

This change is not meant to be permanent.  Rather, it is a pragmatic change that doesn't sacrifice any functionality (we're already mutually excluding calls to `rmw_wait`) but improves behavior until we add the new features regarding management of wait sets that is being worked on in #96. 

This change appears to fix the segfault mentioned in https://github.com/ros2/system_tests/pull/72#issuecomment-157103620.

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_wait.cpp
+++ b/rmw_opensplice_cpp/src/rmw_wait.cpp
@@ -47,7 +47,28 @@ rmw_wait(
   rmw_clients_t * clients,
   rmw_time_t * wait_timeout)
 {
-  DDS::WaitSet waitset;
+  // TODO(gerkey)
+  // We're making this object static to avoid a race between its destructor
+  // being called on exit from this function and an interrupt guard condition
+  // that holds a pointer to it being called in parallel from somewhere else.
+  // This structure will be changed when we add to rmw the feature of supporting
+  // multiple wait sets.
+  static DDS::WaitSet waitset;
+  // To ensure that we properly clean up the now-static wait set, we declare an
+  // object whose destructor will detach what we attached (this was previously
+  // being done inside the destructor of the wait set.
+  struct atexit_t
+  {
+    ~atexit_t()
+    {
+      // Manually detach conditions, to ensure a clean wait set for next time.
+      DDS::ConditionSeq attached_conditions;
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/rmw_wait.cpp
```diff
@@ -47,7 +47,27 @@ rmw_wait(
   rmw_clients_t * clients,
   rmw_time_t * wait_timeout)
 {
-  DDS::WaitSet waitset;
+  // We're making this object static to avoid a race between its destructor
+  // being called on exit from this function and an interrupt guard condition
+  // that holds a pointer to it being called in parallel from somewhere else.
+  // This structure will be changed when we add to rmw the feature of supporting
+  // multiple wait sets.
+  static DDS::WaitSet waitset;
```
dirk-thomas: Should we add a `TODO(...) ...` here?

gerkey:
CI jobs:

* http://ci.ros2.org/job/ci_osx/530/
* http://ci.ros2.org/job/ci_linux/648/
* http://ci.ros2.org/job/ci_windows_connext_static/19/

Other Windows jobs are still in the pipeline.

rmw_opensplice_cpp/src/rmw_wait.cpp
gerkey: 95d99b4

dirk-thomas:
lgtm

gerkey:
Here's the opensplice Windows jobs (the connext ones don't matter for this change):

http://ci.ros2.org/job/ci_windows_opensplice/718/

PR #98, opened 2015-12-02
Keep handle on waitset and condition sequences

jacquelinekay:
Connects to ros2/rmw#49

TODO:
- [x] Implement changes for Connext.
- [ ] Improve error handling in rmw_wait so that the error message of `rmw_fini_waitset` gets returned. goto fail might be a better idea than scope exit here.

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_wait.cpp
+++ b/rmw_opensplice_cpp/src/rmw_wait.cpp
@@ -15,10 +15,13 @@
 #include <ccpp_dds_dcps.h>
 #include <dds_dcps.h>
 
+#include "rmw/allocators.h"
 #include "rmw/error_handling.h"
+#include "rmw/impl/cpp/macros.hpp"
 #include "rmw/rmw.h"
 #include "rmw/types.h"
 
+#include "identifier.hpp"
 #include "types.hpp"
 
 // The extern "C" here enforces that overloading is not used.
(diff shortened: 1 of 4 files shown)


wjwwood:
I think that the synchronization should be done at the highest level (rmw and rcl don't know and shouldn't care if they are being used in a multi-threaded way). We should just make sure to document which functions can and cannot be called in a thread-safe way but I don't think the implementations should try to make them thread-safe. Basically I agree with your point above that `rclcpp`/`rclpy` should enforce that the destructor and the trigger function are not called concurrently.

dirk-thomas:
Linter warning: http://ci.ros2.org/job/ci_linux/672/testReport/rmw_opensplice_cpp/cppcheck/error__uninitvar__src_rmw_waitset_cpp_123_/

jacquelinekay:
Should be fixed by 7dc74dd

rmw_opensplice_cpp/src/rmw_wait.cpp
```diff
@@ -45,31 +48,101 @@ rmw_wait(
...
     {
-      // Manually detach conditions, to ensure a clean wait set for next time.
-      DDS::ConditionSeq attached_conditions;
-      waitset.get_conditions(attached_conditions);
-      for (uint32_t i = 0; i < attached_conditions.length(); ++i) {
-        waitset.detach_condition(attached_conditions[i]);
+      // Manually detach conditions and clear sequences, to ensure a clean wait set for next time.
+      if (!waitset) {
+        RMW_SET_ERROR_MSG("waitset handle is null");
+        return;
```
dirk-thomas: Needs return code.

Same below.

rmw_opensplice_cpp/src/rmw_wait.cpp
```diff
@@ -45,31 +48,101 @@ rmw_wait(
...
+        return )
+      OpenSpliceWaitSetInfo * waitset_info = static_cast<OpenSpliceWaitSetInfo *>(waitset->data);
+      if (!waitset_info) {
+        RMW_SET_ERROR_MSG("WaitSet implementation struct is null");
+        return;
+      }
+
+      DDS::WaitSet * dds_waitset = static_cast<DDS::WaitSet *>(waitset_info->waitset);
+      if (!dds_waitset) {
+        RMW_SET_ERROR_MSG("DDS waitset handle is null");
```
dirk-thomas: Needs return?

Same below.

rmw_opensplice_cpp/src/rmw_wait.cpp
```diff
@@ -45,31 +48,101 @@ rmw_wait(
...
+      }
+
+      for (uint32_t i = 0; i < attached_conditions->length(); ++i) {
+        bool fixed = false;
+        for (uint32_t j = 0; j < waitset->fixed_guard_conditions->guard_condition_count; ++j) {
+          DDS::GuardCondition * fixed_guard_cond =
+            static_cast<DDS::GuardCondition *>(waitset->fixed_guard_conditions->guard_conditions[j
+            ]);
+          if (fixed_guard_cond == (*attached_conditions)[i]) {
+            fixed = true;
```
dirk-thomas: This can `break`.

rmw_opensplice_cpp/src/rmw_wait.cpp
```diff
@@ -45,31 +48,101 @@ rmw_wait(
...
+      if (retcode != DDS::RETCODE_OK) {
+        RMW_SET_ERROR_MSG("Failed to get attached conditions for waitset");
+      }
+
+      for (uint32_t i = 0; i < attached_conditions->length(); ++i) {
+        bool fixed = false;
+        for (uint32_t j = 0; j < waitset->fixed_guard_conditions->guard_condition_count; ++j) {
+          DDS::GuardCondition * fixed_guard_cond =
+            static_cast<DDS::GuardCondition *>(waitset->fixed_guard_conditions->guard_conditions[j
+            ]);
```
dirk-thomas: This wrapping looks weird. Maybe wrap after the open parenthesis?

rmw_opensplice_cpp/src/rmw_wait.cpp
dirk-thomas: Nvm this is in the destructor.

rmw_opensplice_cpp/src/rmw_waitset.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+    if (waitset_info->attached_conditions) {
+      RMW_TRY_DESTRUCTOR_FROM_WITHIN_FAILURE(
+        waitset_info->attached_conditions->~DDS_DCPSUObjSeq(), ConditionSeq)
+      rmw_free(waitset_info->attached_conditions);
+    }
+    if (waitset_info->waitset) {
+      RMW_TRY_DESTRUCTOR_FROM_WITHIN_FAILURE(waitset_info->waitset->~WaitSet(), WaitSet)
+      rmw_free(waitset_info->waitset);
+    }
```
dirk-thomas: If `waitset_info` is not being free here I would propose to set the freed member variables to zero.

dirk-thomas:
The same comments apply to the Connext PR.

rmw_opensplice_cpp/src/rmw_waitset.cpp
jacquelinekay: waitset_info is a pointer to waitset_data, which gets freed on the next line.

rmw_opensplice_cpp/src/rmw_wait.cpp
jacquelinekay: c490d86

rmw_opensplice_cpp/src/rmw_wait.cpp
dirk-thomas: Still applies.

rmw_opensplice_cpp/src/rmw_waitset.cpp
jacquelinekay: Oh, I didn't interpret your original comment correctly. Should I set all freed member variables to zero/null?

jacquelinekay: Fixed.

dirk-thomas:
lgtm.

PR #100, opened 2015-12-04
Use memcpy to avoid breaking strict-aliasing rules

esteve:
This fixes a compiler warning.

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
+++ b/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
@@ -18,6 +18,7 @@
 #include <ccpp_dds_dcps.h>
 #include <u_entity.h>
 
+#include <cstring>
 #include <string>
 
 #include "rosidl_typesupport_opensplice_cpp/impl/error_checking.hpp"
@@ -250,9 +251,11 @@ fail:
   noexcept
   {
     response.sequence_number_ = request_header.sequence_number;
-    response.client_guid_0_ = *(reinterpret_cast<const uint64_t *>(&request_header.writer_guid[0]));
-    response.client_guid_1_ = *(reinterpret_cast<const uint64_t *>(
-        &request_header.writer_guid[0] + sizeof(response.client_guid_0_)));
+    std::memcpy(
+      &response.client_guid_0_, &request_header.writer_guid[0], sizeof(response.client_guid_0_));
+    std::memcpy(
+      &response.client_guid_1_, &request_header.writer_guid[0] + sizeof(response.client_guid_0_),
+      sizeof(response.client_guid_1_));
 
     return TemplateDataWriter<Sample<ResponseT>>::write_sample(response_datawriter_, response);
   }
--- a/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template
+++ b/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template
@@ -9,6 +9,7 @@
 @#  - get_header_filename_from_msg_name (function)
 @#######################################################################
 @
+#include <cstring>
 #include <iostream>
 #include <sstream>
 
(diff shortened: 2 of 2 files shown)


esteve:
Packaging job:

http://ci.ros2.org/view/packaging/job/packaging_linux/56/

CI jobs:

http://ci.ros2.org/job/ci_linux/663/
http://ci.ros2.org/job/ci_osx/546/
http://ci.ros2.org/job/ci_windows_opensplice/738/

wjwwood:
Not sure if you want to do `#include <string.h>` in each of those files, but `memcpy` is technically defined in `string.h`.

esteve:
@wjwwood good catch, I've updated this branch to include `<cstring>` and use `std::memcpy`

dirk-thomas:
Instead of calling `memcpy` this could directly assign the bytes (optionally shifted) as its done in rmw connext: https://github.com/ros2/rmw_connext/blob/master/rmw_connext_dynamic_cpp/src/functions.cpp#L2945-L2947

esteve:
The strict aliasing warnings are fixed now:

http://ci.ros2.org/view/packaging/job/packaging_linux/56/warnings17Result/fixed/

I already fixed the uncrustify and cpplint warnings too, and the rest of the tests pass too.

Should be mergeable.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
```diff
@@ -250,9 +251,11 @@ class Responder
   noexcept
   {
     response.sequence_number_ = request_header.sequence_number;
-    response.client_guid_0_ = *(reinterpret_cast<const uint64_t *>(&request_header.writer_guid[0]));
-    response.client_guid_1_ = *(reinterpret_cast<const uint64_t *>(
-        &request_header.writer_guid[0] + sizeof(response.client_guid_0_)));
+    std::memcpy(
+      &response.client_guid_0_, &request_header.writer_guid[0], sizeof(response.client_guid_0_));
+    std::memcpy(
+      &response.client_guid_1_, &request_header.writer_guid[0] + sizeof(&response.client_guid_0_),
```
gerkey: Is `&request_header.writer_guid[0] + sizeof(&response.client_guid_0_)` a longer way of saying `&request_header.writer_guid[1]`?  Or even `request_header.writer_guid + 1`?

esteve: `writer_guid` is defined as an array of 16 `int8_t` (https://github.com/ros2/rmw/blob/master/rmw/include/rmw/types.h#L100), so both suggestions would point to the second element of the array, instead of the eighth. I'm using `sizeof` instead of hardcoding the array element (8) because I think this looks slightly more readable.

esteve:
I pushed one last change that fixed an incorrect use of `sizeof`, the tests passed locally and on the buildfarm with the incorrect `sizeof`, but I had another look at the code and realized that it was wrong.

New CI jobs:

http://ci.ros2.org/job/ci_linux/668/
http://ci.ros2.org/job/ci_osx/549/
http://ci.ros2.org/job/ci_windows_opensplice/742/

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/responder.hpp
gerkey: Ah, that makes sense.  I hadn't checked the underlying types and just assumed that you were trying to get the 0th and 1st elements, not the 0th-7th and 8th-15th elements.

gerkey:
If CI is happy, then +1 from me.

esteve:
More info about the misuse of `sizeof`. It was a `sizeof` on an address, which by happenstance, is the same size as each client id (`uint64_t`) on x86-64. So in the end the tests passed, but for the wrong reason.

dirk-thomas:
That is why I suggested to not use `memcpy` in the first place but simple assignments. It adds unnecessary complexity and therefore risk to the code

esteve:
@dirk-thomas I don't know, I'm pretty sure I could have screwed it up too :-)

gerkey:
@dirk-thomas Oh, I'd much rather have the complexity of invoking memcpy than the complexity of bit-masking and bit-shifting like in the example you showed.  The memcpy line is way easier to understand.

@esteve Another way to guard against that specific bug with memcpy (taking the size of a pointer) is to always take the size of the type, not the variable. You then have the risk of later changing the type without updating the memcpy line, but that's the same risk as when you're taking the size of a variable when it's an instance of a given type and later having somebody change it to be a pointer to that type.

esteve:
@dirk-thomas in any case, if I understand your suggestion correctly, that's what we already had, which is what causes the strict aliasing rules to break. We used casts and assignments, but when the pointers are dereferenced, that's where the compiler warning happens.

@gerkey yeah, but unfortunately Google's C++ style guide (which is the one we follow), recommends using sizeof on the variable:

https://google.github.io/styleguide/cppguide.html#sizeof

esteve:
CI tests pass, merging.

dirk-thomas:
@esteve The Connext code I referenced does not use pointers at all but only uses values. It is very different from what this code looked before and since it doesn't involve pointers can't have the same warning.

esteve:
@dirk-thomas ah, I see what you meant. Actually, I wrote the code you reference, and I have to be honest, it doesn't look very readable, but at least it's short. In this case, the equivalent code would be 16 lines long and not so readable, so I prefer the `memcpy` version now that all the issues are fixed.

PR #103, opened 2015-12-15
changes to support rcl refactor

wjwwood:
Connects to ros2/rcl#5

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_client.cpp
+++ b/rmw_opensplice_cpp/src/rmw_client.cpp
@@ -37,7 +37,7 @@ rmw_create_client(
   const rmw_node_t * node,
   const rosidl_service_type_support_t * type_support,
   const char * service_name,
-  const rmw_qos_profile_t & qos_profile)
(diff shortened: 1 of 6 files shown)


rmw_opensplice_cpp/src/rmw_client.cpp
```diff
@@ -93,11 +93,11 @@ rmw_create_client(
     goto fail;
   }
 
-  if (!get_datareader_qos(nullptr, qos_profile, datareader_qos)) {
+  if (!get_datareader_qos(nullptr, *qos_profile, datareader_qos)) {
```
jacquelinekay: Same comment that I had in the rmw_connext pull request applies here.

wjwwood:
Same as with Connext, there is a new commit: 3c98216.

rmw_opensplice_cpp/src/rmw_client.cpp
wjwwood: 057ed81

PR #104, opened 2016-01-25
rename message_type_support{_struct}.h

wjwwood:
Connects to ros2/rosidl#71

### DIFF
--- a/rosidl_typesupport_opensplice_c/package.xml
+++ b/rosidl_typesupport_opensplice_c/package.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>rosidl_typesupport_opensplice_c</name>
+  <version>0.0.0</version>
+  <description>Generate the C interfaces for PrismTech OpenSplice.</description>
+  <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>libopensplice64</buildtool_depend>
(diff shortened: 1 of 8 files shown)


jacquelinekay:
+1

jacquelinekay:
I retract my +1, I think I understand what needs to happen now

rosidl_typesupport_opensplice_c/package.xml
```diff
@@ -0,0 +1,35 @@
...
+  <buildtool_depend>rosidl_generator_c</buildtool_depend>
+  <buildtool_depend>rosidl_typesupport_opensplice_cpp</buildtool_depend>
+
+  <buildtool_export_depend>ament_cmake</buildtool_export_depend>
+  <buildtool_export_depend>libopensplice64</buildtool_export_depend>
+  <buildtool_export_depend>opensplice_cmake_module</buildtool_export_depend>
+  <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend>rosidl_generator_c</buildtool_export_depend>
+  <buildtool_export_depend>rosidl_generator_dds_idl</buildtool_export_depend>
+  <!-- The following is needed because the OpenSplice IDL generator is in that package. -->
```
jacquelinekay: Will we eventually move the IDL generator into a separate package that both Opensplice typesupport packages depend on?

jacquelinekay:
the non-isolated build wasn't working because `ament_package` wasn't being called in CMakeLists.txt. fixed in 12fc1f9

+1

wjwwood:
Thanks for the fixup @jacquelinekay, that was another example of me trying to propose part of my changes and failing to include everything that was needed.

PR #109, opened 2016-02-09
Use a set to avoid multiple forward declarations for the same type

esteve:
This prevents repeating a forward declarations when a message contains multiple fields with the same type, e.g. https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/msg/Nested.msg#L3

### DIFF
--- a/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
+++ b/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
@@ -69,10 +69,17 @@ ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
 have_not_included_primitive_arrays = True
 have_not_included_string = True
 }@
-@{
-field_types = set([field.type for field in spec.fields])
+{
+from collections import OrderedDict
+field_types = {}
+for field in spec.fields:
+    if not field.type.is_primitive_type():
+        full_field = field.type.pkg_name + '__msg__' + field.type.type
+        if full_field not in field_types:
+            field_types[full_field] = field.type
+field_types = OrderedDict(sorted(field_types.items()))
 }@
-@[for field_type in field_types]@
+@[for full_field, field_type in field_types.items()]@
 @[if field_type.is_primitive_type()]@
 @[if field_type.is_array and have_not_included_primitive_arrays]@
 @{have_not_included_primitive_arrays = False}@
--- a/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
+++ b/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
@@ -69,30 +69,33 @@ ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
 have_not_included_primitive_arrays = True
 have_not_included_string = True
 }@
-@[for field in spec.fields]@
-@[if field.type.is_primitive_type()]@
-@[if field.type.is_array and have_not_included_primitive_arrays]@
+@{
+field_types = set([field.type for field in spec.fields])
+}@
+@[for field_type in field_types]@
+@[if field_type.is_primitive_type()]@
+@[if field_type.is_array and have_not_included_primitive_arrays]@
 @{have_not_included_primitive_arrays = False}@
 #include <rosidl_generator_c/primitives_array.h>
 #include <rosidl_generator_c/primitives_array_functions.h>
 
 @[end if]@
-@[if field.type.type == 'string' and have_not_included_string]@
+@[if field_type.type == 'string' and have_not_included_string]@
 @{have_not_included_string = False}@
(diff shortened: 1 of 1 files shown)


rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
```diff
@@ -69,30 +69,33 @@ ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
 have_not_included_primitive_arrays = True
 have_not_included_string = True
 }@
-@[for field in spec.fields]@
-@[if field.type.is_primitive_type()]@
-@[if field.type.is_array and have_not_included_primitive_arrays]@
+@{
+field_types = set([field.type for field in spec.fields])
+}@
+@[for field_type in field_types]@
```
tfoote: It would be good to sort the iterated set here for consistency.

tfoote:
one comment otherwise lgtm

esteve:
This code doesn't actually work, each `field.type` has a different hash value, I'm rewriting this now.

esteve:
@tfoote I've changed the code to use an `OrderedDict` and it should iterate over the field in alphabetical order now.

rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
```diff
@@ -69,30 +69,40 @@ ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
...
 }@
-@[for field in spec.fields]@
-@[if field.type.is_primitive_type()]@
-@[if field.type.is_array and have_not_included_primitive_arrays]@
+{
+from collections import OrderedDict
+field_types = {}
+for field in spec.fields:
+    if not field.type.is_primitive_type():
+        full_field = field.type.pkg_name + '__msg__' + field.type.type
```
dirk-thomas: `full_field` seems to be not used anymore. Why does it have to be such a constructed string? `[Base]Type` should be usable as a dictionary key.

Then you also don't need a dictionary in the first place since a list will do just fine.

esteve: Correct me if I'm wrong, but I thought that `[Base]Type` didn't return the package name, so if two messages had the same base type name, but different packages, wouldn't the logic here be incorrect?

dirk-thomas: `BaseType` does contain the packages name: https://github.com/ros2/rosidl/blob/bf8874c8a2286c19590a85e5090fdc43d77cb51f/rosidl_parser/rosidl_parser/__init__.py#L131

esteve: I meant `BaseType.type`. Anyway, it's weird, but I didn't get unique elements in the set, so I thought `BaseType` didn't implement `__hash__`, but I see it does. I've updated the code to use `OrderedSet`

rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.template
```diff
@@ -69,30 +69,34 @@ ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
 have_not_included_primitive_arrays = True
 have_not_included_string = True
 }@
-@[for field in spec.fields]@
-@[if field.type.is_primitive_type()]@
-@[if field.type.is_array and have_not_included_primitive_arrays]@
+{
+from collections import OrderedSet
+field_types = OrderedSet(sorted([field.type for field in spec.fields]))
```
dirk-thomas: Instead of using an ordered set here I would suggest to just collect them in a plain set. Later the include directives should be ordered based on the string value of the included headers to comply with the linter.

jacquelinekay:
We should consider adding a message with repeated fields, like in your example, to other tests suits for typesupport generation. I'll ticket it in rosidl (I bet this is an issue with other message templates).

jacquelinekay:
(Actually, I think is mostly an issue for the middleware typesupport templates, which have forward declaration functions. Therefore the new test messages should be added in test_communication.)

jacquelinekay:
is this PR still needed or should we close it?

tfoote:
@dirk-thomas to check if this is still needed.

dirk-thomas:
This patch is not needed and only a cosmetic improvement (avoiding duplication in generated code). Therefore I will move this back into backlog for now.

Once we are linting the generated code this can be revisited since it will likely complain about the repeated import of the same header files.

dirk-thomas:
This was already addressed in #135. The current master already implements unique and correctly ordered includes (as well as forward declarations): https://github.com/ros2/rmw_opensplice/blob/8e0b144b685eff0f32c9ab5ef66b40c5546d4c44/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.em#L65-L86

So I think this PR is obsolete.

PR #111, opened 2016-02-11
Reset fixed guard condition

gerkey:
Same fix as ros2/rmw_connext#129.

* http://ci.ros2.org/job/ci_linux/946/
* http://ci.ros2.org/job/ci_osx/792/
* http://ci.ros2.org/job/ci_windows/1037 (in the queue)


wjwwood:
lgtm.

It's obviously out of the scope of this pr, but there are other places in these atexit handlers where they set the error string but then just return and are likely being ignored when set.

PR #113, opened 2016-02-12
Only export include directories if installing

esteve:


rosidl_typesupport_opensplice_c/cmake/rosidl_typesupport_opensplice_c_generate_interfaces.cmake
```diff
@@ -284,9 +284,8 @@ if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
...
     RUNTIME DESTINATION bin
   )
-ament_export_libraries(${rosidl_generate_interfaces_TARGET}${_target_suffix} ${OpenSplice_LIBRARIES})
-
-endif()
 
+  ament_export_libraries(${rosidl_generate_interfaces_TARGET}${_target_suffix} ${OpenSplice_LIBRARIES})
 
-ament_export_include_directories(include)
+  ament_export_include_directories(include)
```
dirk-thomas: This should only happen when the package actually installs headers.

Same below.

esteve: So should I remove this? Or put it elsewhere?

dirk-thomas: Look further up where and when headers are being installed to the include folder. The condition is different. It's probably easiest to do it there.

esteve: Thanks, does this look ok? https://github.com/ros2/rmw_opensplice/commit/018e3922102409dd6b51916c2ce034bc33c36c16

dirk-thomas: No, the logical operator needs to be `OR` not `AND`.

esteve: Fixed.

wjwwood:
Can we go ahead and consider this for merge? The changes lgtm and even without the Python stuff, these changes are correct and worth merging, imo.

dirk-thomas:
LGTM

PR #114, opened 2016-02-12
fix inconsistent instance state flags which results in a busy loop

dirk-thomas:
Based on the investigations here: https://github.com/ros2/rclcpp/issues/192#issuecomment-183491185

It makes the test from the `issue_192` branch work for me locally.

* http://ci.ros2.org/job/ci_linux/975/
* http://ci.ros2.org/job/ci_osx/810/
* http://ci.ros2.org/job/ci_windows/1063/

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template
+++ b/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template
@@ -88,6 +88,8 @@ public:
       1,
       DDS::ANY_SAMPLE_STATE,
       DDS::ANY_VIEW_STATE,
+      // See ros2/rclcpp#192 for a justification of this option:
+      // https://github.com/ros2/rclcpp/issues/192#issuecomment-183491185
       DDS::ANY_INSTANCE_STATE);
 
     switch (status) {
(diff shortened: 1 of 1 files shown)


wjwwood:
I'd recommend adding a detailed comment about why this is the way it is. I feel that in a year we could easily look at this and think it should be another way, having forgotten why we made this change. I was curious to see if we had an interesting commit message when we added that option, but the file has been refactored and renamed, which makes it hard to follow back to the original commit.

dirk-thomas:
I traced both locations back to its origin and they have both been added with exactly these arguments when the features of services and then later the read conditions have been added.

wjwwood:
I did too. I think a comment in the code makes sense because even though I could trace it back, it was pretty hard and very time consuming.

dirk-thomas:
Please feel free to suggest a comment and add it to the PR.

dirk-thomas:
Since this change doesn't have any regressions on Linux and OS X I would like to merge this now (even with the Windows job still pending, but that will be the case for at least another hour). Any +1 votes?

jacquelinekay:
I agree with William. I will add a comment and then approve.

jacquelinekay:
+1

wjwwood:
lgtm

PR #118, opened 2016-02-24
Set topic names

jacquelinekay:
Connects to ros2/rcl#29

Without this PR, the test in the above PR to rcl fails for opensplice.

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_publisher.cpp
+++ b/rmw_opensplice_cpp/src/rmw_publisher.cpp
@@ -170,6 +170,12 @@ rmw_create_publisher(
 
   publisher->implementation_identifier = opensplice_cpp_identifier;
   publisher->data = publisher_info;
+  publisher->topic_name = reinterpret_cast<const char *>(rmw_allocate(strlen(topic_name) + 1));
+  if (!publisher->topic_name) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for node name");
+    goto fail;
+  }
+  memcpy(const_cast<char *>(publisher->topic_name), topic_name, strlen(topic_name) + 1);
 
   return publisher;
 fail:
@@ -260,6 +266,9 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
       }
     }
     rmw_free(publisher_info);
+  }
+  if (publisher->topic_name) {
+    rmw_free(const_cast<char *>(publisher->topic_name));
   }
   rmw_publisher_free(publisher);
   return result;
--- a/rmw_opensplice_cpp/src/rmw_subscription.cpp
+++ b/rmw_opensplice_cpp/src/rmw_subscription.cpp
@@ -173,6 +173,14 @@ rmw_create_subscription(
 
   subscription->implementation_identifier = opensplice_cpp_identifier;
   subscription->data = subscriber_info;
+
+  subscription->topic_name = reinterpret_cast<const char *>(rmw_allocate(strlen(topic_name) + 1));
+  if (!subscription->topic_name) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for node name");
+    goto fail;
+  }
+  memcpy(const_cast<char *>(subscription->topic_name), topic_name, strlen(topic_name) + 1);
+
   return subscription;
 fail:
   if (dds_subscriber) {
@@ -282,6 +290,9 @@ rmw_destroy_subscription(rmw_node_t * node, rmw_subscription_t * subscription)
       }
     }
     rmw_free(subscription_info);
+  }
+  if (subscription->topic_name) {
+    rmw_free(const_cast<char *>(subscription->topic_name));
   }
   rmw_subscription_free(subscription);
   return result;


wjwwood:
Is it sufficient to just assign it?

In node it has space allocated for it and it is copied:

https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/src/rmw_node.cpp#L178-L183

And destroyed:

https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/src/rmw_node.cpp#L228-L230
https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/src/rmw_node.cpp#L292-L293

jacquelinekay:
Good point, it's better practice to copy the string. I guess that's the complication you were just talking about.

rmw_opensplice_cpp/src/rmw_publisher.cpp
```diff
@@ -170,6 +170,13 @@ rmw_create_publisher(
 
   publisher->implementation_identifier = opensplice_cpp_identifier;
   publisher->data = publisher_info;
+  publisher->topic_name = reinterpret_cast<const char *>(
+    rmw_allocate(sizeof(char) * strlen(topic_name) + 1));
```
dirk-thomas: I don't think `sizeof(char)` is needed her but when it is used the math is wrong: `sizeof(topic_name)` and ` + 1` needs to be in parenthesis.

rmw_opensplice_cpp/src/rmw_publisher.cpp
```diff
@@ -170,6 +170,13 @@ rmw_create_publisher(
...
+  if (!publisher->topic_name) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for node name");
+    goto fail;
+  }
+  memcpy(const_cast<char *>(publisher->topic_name), topic_name, strlen(topic_name) + 1);
```
dirk-thomas: Should we use `strcpy` / `strncpy` here?

rmw_opensplice_cpp/src/rmw_publisher.cpp
jacquelinekay: True.

`sizeof(char)` will evaluate to 1 on most platforms which is why it doesn't make a difference. But for completeness parentheses will be added. I will fix it here and in the place I copy-pasted it from :)

dirk-thomas: For platform where `sizeof(char)` is not `1` the memcpy code below will not work anyway.

rmw_opensplice_cpp/src/rmw_publisher.cpp
wjwwood: I usually just default to `memcpy` incase I'm dealing with binary data in the string, but in this case `strcpy` (or `strncpy` to avoid stupid warnings on Windows) would be fine since we're also using `strlen`.

rmw_opensplice_cpp/src/rmw_publisher.cpp
wjwwood: I use `sizeof(char)` in combination with `malloc`-ish functions as a way to express I'm creating space for `char`'s. If you just use `1` it isn't very obvious what you're doing, and leaving out the `sizeof` is a bad habit which might show up when you really should be using it. I'd say leave the `sizeof(char)` and fix the `memcpy` to include it if you like.

rmw_opensplice_cpp/src/rmw_publisher.cpp
jacquelinekay: Regarding the above discussion about multiplying by sizeof(char), it seems like if we switch to using strcpy down here it makes sense to leave out to sizeof(char) here as well (but to keep it in the above rmw_allocate call).

I don't know much about platforms where sizeof(char) != 1 (seems very rare/nonstandard?) but it seems like `str` functions would be implemented with that in mind and the unit for the len argument for `strn` functions would account for that...

jacquelinekay: Maybe I'll do some research.

jacquelinekay: http://stackoverflow.com/questions/437470/type-to-use-to-represent-a-byte-in-ansi-c89-90-c/437640#437640

"char is always a byte, but it's not always an octet."
"sizeof(char) is always 1 for all implementations"

wjwwood: I don't really care what we do when copying -- we can use memcpy with or without `sizeof(char)` or use `strncpy` + `strlen`. For the `memcpy` call, I was just trying to say the `sizeof(char)` is for readability not correctness or trying to support systems where `char` is not a byte.

wjwwood: Sorry that should have been "For the `malloc`-like call, ...". I updated my previous comment.

jacquelinekay: In light of this discovery, I'm going to take out sizeof(char) here.

jacquelinekay: I realized why we don't use strncpy everywhere, it's because it provokes a warning on Windows. Yarr. http://ci.ros2.org/job/ci_windows/1110/warnings34Result/new/package.-203766663/source.6589365479437991052/#1133

wjwwood:
lgtm if the warnings were resolved.

jacquelinekay:
I'm going to run CI on Windows with Opensplice only, juuuust in case

jacquelinekay:
http://ci.ros2.org/job/ci_windows/1155/

jacquelinekay:
actually, it looks like I retriggered a job and posted it on the thread for ros2/rcl#29, and it looks like these should be fine to merge

PR #122, opened 2016-03-24
defer loading OpenSplice dependency until this package is found

dirk-thomas:
See ros2/ros2#207.

~~http://ci.ros2.org/view/packaging/job/packaging_windows/175/~~

**Update:** the Windows job is failing - I am looking into it...

### DIFF
--- a/rmw_opensplice_cpp/CMakeLists.txt
+++ b/rmw_opensplice_cpp/CMakeLists.txt
@@ -27,14 +27,12 @@ endif()
 find_package(rmw REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
 
-ament_export_definitions(${OpenSplice_DEFINITIONS})
 ament_export_dependencies(
   rmw
   rosidl_generator_cpp
   rosidl_typesupport_opensplice_c
   rosidl_typesupport_opensplice_cpp
 )
-ament_export_include_directories(${OpenSplice_INCLUDE_DIRS})
 
 link_directories(${OpenSplice_LIBRARY_DIRS})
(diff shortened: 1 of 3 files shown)


dirk-thomas:
http://ci.ros2.org/job/ci_linux/1111/
http://ci.ros2.org/job/ci_osx/892/
http://ci.ros2.org/job/ci_windows/1178/

http://ci.ros2.org/view/packaging/job/packaging_windows/177/

wjwwood:
Looks like http://ci.ros2.org/job/ci_linux/1111/console failed. Is that a race condition from a missing make dependency? I don't think this is caused by this pr, but I can't say for certain that it isn't related. Should we run another CI job?

dirk-thomas:
I already ran another job and updated the link in the previous comment to point to build number 1112.

The build is failing due to the Python code generator. I saw in other builds that the Python generator is triggered multiple times within a single package when building with multiple threads (look for `Generating Python code for ROS interfaces` in the console output). I am currently looking into it locally but something is very buggy. I assume the race condition is just another symptom of the same underlying problem.

dirk-thomas:
The Python generator problem was unrelated and was addresses in ros2/rosidl#111.

wjwwood:
The code changes lgtm. I expressed my concern that this doesn't completely make it possible to relocate OpenSplice on the target computer: https://github.com/ros2/ros2/issues/207#issuecomment-201078879 (which I believe is the goal of this pr)

dirk-thomas:
Since this PR only aiming at the CMake part of the problem I will go ahead and merge it. For full relocation ability the environment hook also need some work.

PR #125, opened 2016-03-29
Implement rosnode and rostopic functionality

vmayoral:


rmw_opensplice_cpp/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ endif()
 
 find_package(rmw REQUIRED)
 find_package(rosidl_generator_cpp REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(introspection_msgs REQUIRED)
```
wjwwood: I think this is going to be a problem with this approach. I do not believe it makes sense to have our rmw implementations depend on and use our messages.

Even though it might be more work, I think the right thing to do is to figure out how to expose the needed information up through the rmw interface and then any work that needs to publish or subscribe to topics should be done in `rcl`.

I'll have to see what all is happening under the hood in order to make a more concrete suggestion.

rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -294,4 +387,180 @@ rmw_destroy_node(rmw_node_t * node)
...
+    RMW_SET_ERROR_MSG("environment variable ROS_DOMAIN_ID is not set");
+    fprintf(stderr, "[rmw_opensplice_cpp]: error: %s\n", rmw_get_error_string_safe());
+    return nullptr;
+  }
+#endif
+
+  // Create a ROS middleware node that will be used to 
+  // fetch other participant's data from the "ros_meta" topic
+  rmw_node_t * rmw_node;
+  rmw_node = rmw_create_node("get_node_names", domain_id);
```
wjwwood: I think we should avoid creating an internal node just to get the metadata. I think it would be better to have each node do this even if that's less efficient in the short term. We can work on avoiding duplicated effort within a single process later by using a daemon or some intelligent shared resource.

dirk-thomas:
This functionality has been implemented on top of the rmw layer in `ros2cli` by now.

PR #130, opened 2016-05-02
also search for opensplice in ament prefix path

dirk-thomas:
Allows to build opensplice within an ament workspace (currently necessary on platforms we don't have Debian packages for, e.g. Xenial).

Depends on ament/ament_tools#93.


tfoote:
+1

dirk-thomas:
Can you please clarify if the +1 is only for this patch or also for the recursive tickets ament/ament_tools#93 and ament/ament_package#39.

tfoote:
reviewed the recursive tickets too

PR #131, opened 2016-05-06
fixed null terminator check

mikaelarguedas:
This fixes the null terminator character check in strings

Connects to https://github.com/ros2/rclpy/issues/17


wjwwood:
+1

PR #136, opened 2016-05-18
Don't use C++ typesupport for C services

jacquelinekay:
similar approach to ros2/rmw_connext#164


jacquelinekay:
CI:
http://ci.ros2.org/job/ci_linux/1334/
http://ci.ros2.org/job/ci_osx/1047/
http://ci.ros2.org/job/ci_windows/1351/

jacquelinekay:
http://ci.ros2.org/job/ci_osx/1053/badge/icon Hopefully the OSX warnings will be gone here.

jacquelinekay:
Warnings are fixed; there are some FastRTPS services failures although I think these occur on the nightly builds as well.

wjwwood:
I believe (this is also the case for the connext pr), that the critical changes are in the `rosidl_typesupport_opensplice_c/resource/srv__type_support_c.cpp.em` file and the all the other changes are just to support this line of code:

```c++
@(spec.pkg_name)__srv__@(spec.srv_name)_Response__convert_dds_to_ros(
      static_cast<void *>(&response.data()), untyped_ros_response);
```

I think that could also be accomplished with a much less intrusive change by doing the above snippet as something like this:

```c++
const rosidl_message_type_support_t * ts =
  ROSIDL_GET_TYPE_SUPPORT(@(spec.pkg_name), srv, @(spec.srv_name));
const message_type_support_callbacks_t * callbacks =
  static_cast<const message_type_support_callbacks_t *>(ts->data);
callbacks->convert_dds_to_ros(
  static_cast<void *>(&response.data()), untyped_ros_response);
```

This is admittedly a little more work, in the service functions, but it matches the way it works in `rmw_publish` for example and is a little more dogfood-ish (is that a word?).

What do you think about changing this pr to do that instead? I'm happy to author the cleanup to save you time.

jacquelinekay:
Yeah, sounds like a much better design. I'm cleaning up the rcl parameters pull request right now (better error handling and cleanup on error) so it would be great if you could handle the changes.

wjwwood:
Ok, I'll start with this pr first.

wjwwood:
I did "rebuild" on the CI you posted in https://github.com/ros2/rmw_opensplice/pull/136#issuecomment-222000006:

- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1062/)

jacquelinekay:
Thank you!

jacquelinekay:
It seems Jenkins fails, as does my local build.

wjwwood:
Yeah, I'm fixing it right now. Need to get the declaration for the "get type support" function.

wjwwood:
Took me a minute to figure out the right fix: a6665cb

I'll start CI when my local gets past a few key packages.

jacquelinekay:
I'll build it locally too with C parameters.

rosidl_typesupport_opensplice_c/resource/srv__type_support_c.cpp.em
```diff
@@ -29,8 +29,6 @@
 #include "@(spec.pkg_name)/srv/dds_opensplice/ccpp_Sample_@(spec.srv_name)_Response_.h"
 #include "rosidl_typesupport_opensplice_c/identifier.h"
 
-#include "@(spec.pkg_name)/srv/dds_opensplice_c/@(get_header_filename_from_msg_name(spec.srv_name))__request__type_support.h"
-#include "@(spec.pkg_name)/srv/dds_opensplice_c/@(get_header_filename_from_msg_name(spec.srv_name))__response__type_support.h"
 #include "@(spec.pkg_name)/srv/dds_opensplice/@(get_header_filename_from_msg_name(spec.srv_name))__type_support.cpp"
```
jacquelinekay: @wjwwood testing this locally, I get a segfault in the tests I have for C parameters. I'm not entirely sure, but I think this needs to be changed to "__type_support_c.cpp". Will do more testing and push a fix if I find one.

wjwwood:
I'm working on a compiler error fix in the message template.

wjwwood:
This branch is ok.

wjwwood:
New complete CI check:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1361/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1063/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1364/)

wjwwood:
Looks like there are some linker warnings on Windows: http://ci.ros2.org/job/ci_windows/1364/warnings34Result/new/package.465770450/file.199757930/

Not sure if they're caused by this pr or not, but it seems likely. I haven't had a chance to look into them yet.

jacquelinekay:
They do appear to be caused by this PR.

jacquelinekay:
Actually, looking more carefully at the source of the warnings and my local build, the warnings appear to be due to the Connext branch with the same name. Relaunching this branch only under the name `fix_cpp_convert_ospl`:

[Build Status](http://ci.ros2.org/job/ci_windows/1367/)

wjwwood:
It does look as if those warnings are gone with OpenSplice only. I'll merge the OpenSplice branch and you can continue to look into the Connext problems.

jacquelinekay:
It does look like this test may have introduced 10 new failing tests on Windows though, comparing it to nightly: http://ci.ros2.org/job/ci_windows/1367/

wjwwood:
The only OpenSplice related test that failed was http://ci.ros2.org/job/ci_windows/1367/testReport/(root)/projectroot/gtest_executor__rmw_opensplice_cpp/, which I think is flaky. The others are `rmw_connext_cpp` tests. That might be due to my changes. I'll look into it.

jacquelinekay:
I just noticed that this wasn't squashed before merging, was that intentional?

wjwwood:
Nope, I thought I used the "squash and commit" option in GitHub's merge button.

wjwwood:
Hmm, I was sure I checked that... That's frustrating, sorry.

jacquelinekay:
It's fine

tfoote:
I'm not sure why squash and merge is default on some repos but not others. There does not seem to be a setting for which should be the default button.

wjwwood:
@tfoote yeah, I've noticed that and that's why I always check it. I must have misclicked or done it too quickly or something.

dirk-thomas:
Might these changes be responsible for the new warnings in the Windows packaging job: http://ci.ros2.org/job/packaging_windows/253/warnings34Result/

PR #137, opened 2016-05-20
support wait_for_service

wjwwood:
Connects to ros2/ros2#215

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
+++ b/rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
@@ -26,7 +26,6 @@
 // The extern "C" here enforces that overloading is not used.
 extern "C"
 {
-
 rmw_ret_t
 rmw_service_server_is_available(
   const rmw_node_t * node,
@@ -72,5 +71,4 @@ rmw_service_server_is_available(
 
   return RMW_RET_OK;
 }
-
 }  // extern "C"
--- a/rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
+++ b/rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
@@ -1,0 +1,76 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "rmw/allocators.h"
+#include "rmw/error_handling.h"
+#include "rmw/impl/cpp/macros.hpp"
+#include "rmw/rmw.h"
+
+#include "identifier.hpp"
+#include "types.hpp"
+
+// The extern "C" here enforces that overloading is not used.
+extern "C"
```
jacquelinekay: do you need to wrap this with `#if __cplusplus` here or is that only needed in header files?

jacquelinekay: Yeah nevermind, I don't think it's needed.

wjwwood:
This is also ready for review. The latest commit 6f65a4e fixed the bug I was struggling with where the publisher and subscription count was always going up and never down.

rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
```diff
@@ -282,6 +285,46 @@ class Requester
...
+  /// Return NULL on success, otherwise an error string.
+  const char *
+  server_is_available(
+    const rmw_node_t * node, bool * is_available,
+    rmw_ret_t (* count_publishers)(const rmw_node_t *, const char *, size_t *),
+    rmw_ret_t (* count_subscribers)(const rmw_node_t *, const char *, size_t *)
+  ) noexcept
+  {
+    if (!is_available) {
+      return "argument is_available is null";
```
tfoote: should be return nullptr with setting error message?

wjwwood: No, these functions are strange like that. See: https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/service_type_support.h#L62-L66

tfoote:
lgtm, there are several error catches with fprintfs such that a user cannot deal with them. Most are clean up errors. Presumeably if we cannot deallocate things there's something bad happening.  I don't think we have a way to expose that sort of failure at the moment though.

wjwwood:
Yeah, I try to only use `fprintf` when returning an error is not possible, like in the "fail" goto label of a create function.

PR #139, opened 2016-05-31
Fix warnings on Windows

jacquelinekay:
This should fix the warnings on Windows for generated Opensplice C messages.

### DIFF
--- a/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.em
+++ b/rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.em
@@ -61,11 +61,6 @@ extern "C"
 {
 #endif
 
-// Forward declare the get type support function for this type.
-ROSIDL_GENERATOR_C_EXPORT_@(spec.base_type.pkg_name)
-const rosidl_message_type_support_t *
-  ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
-
 // include message dependencies
 @{
 includes = {}
--- a/rosidl_typesupport_opensplice_c/cmake/rosidl_typesupport_opensplice_c_generate_interfaces.cmake
+++ b/rosidl_typesupport_opensplice_c/cmake/rosidl_typesupport_opensplice_c_generate_interfaces.cmake
@@ -220,7 +220,7 @@ if(WIN32)
   target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
     PRIVATE "ROSIDL_TYPESUPPORT_OPENSPLICE_C_BUILDING_DLL")
   target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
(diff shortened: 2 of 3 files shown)


jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_windows/1369/) (Opensplice only)

rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.em
```diff
@@ -61,11 +61,6 @@ extern "C"
 {
 #endif
 
-// Forward declare the get type support function for this type.
-ROSIDL_GENERATOR_C_EXPORT_@(spec.base_type.pkg_name)
-const rosidl_message_type_support_t *
-  ROSIDL_GET_TYPE_SUPPORT_FUNCTION(@(pkg), @(subfolder), @(msg))();
-
```
wjwwood: After removing this, do you not need to include the generated header that declares this function for you (with the right dll linkage)?

jacquelinekay: I thought that would get pulled in by `rosidl_generator_c/message_type_support.h`:

https://github.com/ros2/rmw_opensplice/pull/139/files/938fb9a33a5783053cc1053c625596014f53e39b#diff-eb52319d09d2b6415f95c504f6609ba7L29

But now I'm not so sure, I was wrong when I initially thought this got rid of all warnings.

wjwwood: Unfortunately that only defines the `ROSIDL_GET_TYPE_SUPPORT_FUNCTION` macro, but does not use it to declare the message specific version of the function. That is done here:

https://github.com/ros2/rosidl/blob/d09b0edf92d3bc5cf05974fbff20a78ee1ddf70a/rosidl_generator_c/resource/msg__type_support.h.em#L46-L49

jacquelinekay: It seems to compile without including that header directly--maybe it gets pulled in through one of the other rosidl_generator_c includes? I can add an explicit include for that file however.

jacquelinekay:
I got the BUILDING_DLL symbol name wrong the first time.

[Build Status](http://ci.ros2.org/job/ci_windows/1371/)

rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.em
wjwwood: Dunno, I'm fine as long as it works. Though it would be nice to know why it works :).

jacquelinekay: Yeah... I looked into it a bit. My original idea that the `type_support.h` file gets included by one of the other headers looks like it's not correct. It might work because there is still a forward declaration for ROSIDL_GET_TYPE_SUPPORT_FUNCTION for message subfields lower in the file:

Issue #99, opened 2015-12-03
Handle Server runs out of handle space


jacquelinekay:
Originally discovered while testing ros2/rclcpp#165

The easiest way to reproduce this is to choose any gtest that exercises the client library using `rmw_opensplice_cpp` with the command line argument `--gtest_repeat=-1`, which will repeat the test indefinitely.

For example: `./gtest_executor__rmw_opensplice_cpp --gtest_repeat=-1`

After over 100 to several hundred repeated iterations (depending on the machine), the process will exit with the following error:

```
========================================================================================
Report      : ERROR
Date        : Thu Dec 03 09:53:20 PST 2015
Description : The Handle Server ran out of handle space
Node        : pol
Process     : gtest_executor__rmw_opensplice_cpp <24792>
Thread      : main thread 7fde172d1780
Internals   : V6.4.140407OSS///kernel::v_handle::v_handleServerRegister/v_handle.c/235/0/1449165200.978805977
```

To me, this suggests that we need to improve our teardown of middleware handles within rmw.

(I am investigating if there is a similar problem in Connext now.)

gerkey:
I don't know whether it's relevant, but I got this error after ~100 iterations of that test in a Debug build of ROS 2.

jacquelinekay:
I'm running the same test case used in the example in the description with Connext and I'm on iteration 900 now. I don't think this is a problem in Connext, unless their resource storage is significantly better than Opensplice.

gerkey:
Possibly related are crashes that we're seeing with repeated invocations of certain tests (e.g., http://ci.ros2.org/job/ci_windows/1417/testReport/%28root%29/projectroot/gtest_executor__rmw_opensplice_cpp/). Below is a backtrace showing that a test segfaults while trying to create a node. This crash seems to go away if you comment out all the other test cases in that file. That makes me think that OpenSplice isn't being correctly cleaned up when used repeatedly in a single process.

```
(gdb) bt
#0  0x00007f12f003beb7 in mprotect () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007f12f0cf453a in allocate_stack (stack=<synthetic pointer>, pdp=<synthetic pointer>, attr=<optimized out>)
    at allocatestack.c:649
#2  __pthread_create_2_1 (newthread=0x216fad0, attr=0x7fff91f63180, start_routine=0x7f12efcae4a0, arg=0x214e930)
    at pthread_create.c:539
[...]

PR #141, opened 2016-06-07
fix returning loan

dirk-thomas:
As suggested by @wjwwood in https://github.com/ros2/examples/issues/102#issuecomment-224172321

I can confirm that it fixes the problem mentioned in https://github.com/ros2/examples/issues/102#issuecomment-222859255

Running CI for this might not make much sense since it won't show if the patch has the intended effect.


dirk-thomas:
I can confirm that it fixes the problem mentioned in https://github.com/ros2/examples/issues/102#issuecomment-222859255

Running CI for this might not make much sense since it won't show if the patch has the intended effect.

wjwwood:
+1

gerkey:
I wonder whether this fix might help with the cleanup problems identified in #99.

tfoote:
I just pulled and am rebuilding in debug to test #99 and will see if it fixes it.

dirk-thomas:
#99 is still happening for me after 166 iterations.

PR #143, opened 2016-06-09
changes to support wait for service (round 2)

wjwwood:
The important change in this pr is that it adds mutex locking around access to the data structures where we store discovery information. For our OpenSplice implementation, the thread in which the data is added/removed is always different then thread that is asking for the information through `rmw_count_{publishers, subscribers}` and `rmw_get_topic_names_and_types`.

The other change was to introduce some extra console logging of discovery activity with the `RMW_PRINT_DISCOVERY_LOGGING` env variable.

Connects to ros2/ros2#215

### DIFF
--- a/rmw_opensplice_cpp/src/types.cpp
+++ b/rmw_opensplice_cpp/src/types.cpp
@@ -27,6 +27,30 @@ create_type_name(
          "::" + sep + "::dds_::" + callbacks->message_name + "_";
 }
 
+size_t
+CustomDataReaderListener::count_topic(const char * topic_name)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto it = topic_names_and_types_.find(topic_name);
+  if (it == topic_names_and_types_.end()) {
+    return 0;
+  } else {
+    return it->second.size();
+  }
+}
+
+void
+CustomDataReaderListener::fill_topic_names_and_types(
+  std::map<std::string, std::set<std::string>> & tnat)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  for (auto it : topic_names_and_types_) {
+    for (auto & jt : it.second) {
+      tnat[it.first].insert(jt);
+    }
+  }
+}
+
 void
 CustomDataReaderListener::add_information(
   const DDS::SampleInfo & sample_info,
@@ -34,29 +58,29 @@ CustomDataReaderListener::add_information(
   const std::string & type_name)
 {
   // store topic name and type name
-  auto & topic_types = topic_names_and_types[topic_name];
+  auto & topic_types = topic_names_and_types_[topic_name];
   topic_types.insert(type_name);
   // store mapping to instance handle
   TopicDescriptor topic_descriptor;
   topic_descriptor.instance_handle = sample_info.instance_handle;
   topic_descriptor.name = topic_name;
   topic_descriptor.type = type_name;
-  topic_descriptors.push_back(topic_descriptor);
+  topic_descriptors_.push_back(topic_descriptor);
 }
 
 void
 CustomDataReaderListener::remove_information(const DDS::SampleInfo & sample_info)
 {
   // find entry by instance handle
-  for (auto it = topic_descriptors.begin(); it != topic_descriptors.end(); ++it) {
+  for (auto it = topic_descriptors_.begin(); it != topic_descriptors_.end(); ++it) {
     if (it->instance_handle == sample_info.instance_handle) {
       // remove entries
-      auto & topic_types = topic_names_and_types[it->name];
+      auto & topic_types = topic_names_and_types_[it->name];
       topic_types.erase(topic_types.find(it->type));
       if (topic_types.empty()) {
-        topic_names_and_types.erase(it->name);
+        topic_names_and_types_.erase(it->name);
       }
-      topic_descriptors.erase(it);
+      topic_descriptors_.erase(it);
       break;
     }
   }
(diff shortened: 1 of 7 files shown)


wjwwood:
The other thing this does (that I forgot to mention) is that it creates the participant in a disabled state, add the listeners, and then enable the participant. This is supposed to prevent us from missing discovery information in the period of time between creating the participant and attaching the listeners to the builtin datareaders.

I think this is the "right" thing to do, however, I don't know if it's actually helping us in OpenSplice since it doesn't seem to improve our ability to quickly detect existing entities on startup of a new node. It might be that OpenSplice just doesn't do this right, see:

https://groups.google.com/forum/#!topic/ospl-developer/Hg9DxsgW_v0

@codebot and I spent some time this afternoon debugging it, and our best theory is that the above mailing list thread never got resolved and that's causing our troubles. Though we stopped short of looking in detail at the transactions in Wireshark. We could spend more time debugging this problem, but I wanted to get this up for review in the meantime.

I'm going to rewrite the history to have the "create in a disabled state" changes in a separate commit.

wjwwood:
I just pushed 1a73e14 which also fixes a bug I noticed where a server takes a bogus service call when a client disconnects. I observed this:

```
% /Users/william/ros2_ws/build_isolated/rclcpp_examples/add_two_ints_server__rmw_opensplice_cpp test_executable
Incoming request
a: 2 b: 3
Incoming request
a: 140734592188415 b: 140710532199664
```

Where I only ran the client once. After looking at the code, it's obvious that it would mark "taken" true even if the sample info's `valid_data` was `false`.

rmw_opensplice_cpp/src/types.cpp
```diff
@@ -27,36 +31,120 @@ create_type_name(
...
+}
+
+size_t
+CustomDataReaderListener::count_topic(const char * topic_name)
+{
+  std::lock_guard<std::mutex> lock(mutex_);
+  auto it = topic_names_and_types_.find(topic_name);
+  if (it == topic_names_and_types_.end()) {
+    return 0;
+  } else {
```
jacquelinekay: nitpick, you can omit the "else" and just `return it->second.size();` at the end of this function.

rmw_opensplice_cpp/src/types.cpp
```diff
@@ -27,36 +31,120 @@ create_type_name(
...
+void
+print_discovery_logging(
+  const std::string & prefix,
+  const std::string & name,
+  const std::string & type,
+  CustomDataReaderListener::EndPointType end_point_type)
+{
+  // filter builtin OpenSplice topics
+  if (
+    name.compare(0, 4, "DCPS") != 0 &&
```
jacquelinekay: We could potentially receive logging information from other topics if they begin with these tokens, right? I guess the ROS 2 scheme for naming DDS topics will prevent ROS 2 publishers/subscribers from "impersonating" builtin Opensplice topics, which just leaves other DDS implementations--which we probably can't do very much about...

rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.em
```diff
@@ -146,14 +149,12 @@ public:
                "the data_values and info_seq were not obtained from this "
                "@(__dds_sample_type_prefix)@(suffix)_DataReader";
       case DDS::RETCODE_OK:
-        *taken = true;
-        return nullptr;
+        break;
```
jacquelinekay: Any reason for changing from return to break here? It looks like it effectively makes no difference, I'm just curious.

jacquelinekay:
+1, I have a few small comments but overall it makes sense to me.

rmw_opensplice_cpp/src/types.cpp
wjwwood: Yeah... These are just topics that always exist with OpenSplice and there are about 15 of them, so it makes it really hard to find signal in the noise when you don't filter these. We do run the risk of hiding other topics that we are actually looking for which also start with these prefixes, but if that becomes a problem we can adjust this again.

rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.em
wjwwood: I believe we need it, otherwise it would go into the default case.

wjwwood: Oh you mean that nothing else happens after the switch, so we might as well return nullptr here. I did it because there is another spot with similar code that does do extra work after the switch statement but before returning. There's no particular reason. Maybe so someone doesn't add stuff after the switch but before the return and not notice the function returns here.

rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.em
```diff
@@ -124,6 +124,9 @@ public:
 
     if (sample_infos.length() > 0 && sample_infos[0].valid_data) {
       sample = reinterpret_cast<Sample<@(__dds_msg_type_prefix)@(suffix)_> &>(dds_messages[0]);
+      *taken = true;
+    } else {
+      *taken = false;
```
dirk-thomas: This could be restructured like this:

```
*taken = sample_infos.length() > 0 && sample_infos[0].valid_data
if(*taken) {
  sample = ...;
}

rmw_opensplice_cpp/src/types.cpp
wjwwood: dde464b

rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.em
wjwwood: 5a60e4f

rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.em
jacquelinekay: Yeah, my thought that was in case someone decides to follow the same pattern as the other function this might be better. Whatever, either one is fine.

PR #146, opened 2016-06-15
require CMake 3.5

dirk-thomas:
Connect to ament/ament_cmake#72


dirk-thomas:
The removed dependency was invalid since the specified target doesn't exist. In CMake 3.5 that fails where as before it only resulted in a warning. The library target doesn't need a dependency on anything related to idl files anyway. The library target already depend on the source files and the generator for the source files depends on all idl files to be present.

wjwwood:
I don't think removing it (c11110e) is the right fix since it also exists in the cpp version of the file (https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/cmake/rosidl_typesupport_opensplice_cpp_generate_interfaces.cmake#L21), but if you think it isn't necessary then you should remove the cpp instance too.

wjwwood:
Ah, I see what you mean about the dependency being implicit. I still think the cpp equivalent snippet should be removed for consistency.

dirk-thomas:
Can you please point to the exact lines which should be removed from the cpp instance.

wjwwood:
https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/cmake/rosidl_typesupport_opensplice_cpp_generate_interfaces.cmake#L182-L185

dirk-thomas:
That line does the opposite. It add a dependency to the idl target. That is not the same and can't be removed.

wjwwood:
I see now, ok nevermind then I was wrong. lgtm

PR #150, opened 2016-09-13
Add warning for empty OSPL_HOME variable

Karsten1987:
warn the user before setting default location for opensplice

### DIFF
--- a/opensplice_cmake_module/env_hook/opensplice.bat.in
+++ b/opensplice_cmake_module/env_hook/opensplice.bat.in
@@ -1,4 +1,7 @@
-set "OSPL_HOME=@OpenSplice_HOME@"
+if "%OSPL_HOME%" == "" (
+  set "OSPL_HOME=@OpenSplice_HOME@"	
+  echo OSPL_HOME variable not set. Assuming %OSPL_HOME%
+)
 
 call "%OSPL_HOME%\release.bat" 1> nul
 


Karsten1987:
see https://github.com/ros2/ros2/issues/268

wjwwood:
Mmm Batch, lgtm.

dirk-thomas:
Maybe we should make the logic consistent with Connext. Either change this to warn about the different path and overwrite it with the path it was compiled with. Or update Connext to stick to the sourced environment rather then overwriting it with the one from configure time.

Karsten1987:
imho we should use the environment variable. Otherwise we would force everybody to install opensplice (and maybe other deps) in the exact path as the release was configured.

dirk-thomas:
I agree, that makes sense. But I think there should still be a message to let users know about the difference. Otherwise they might accidentally use a different non ABI compatible version

Karsten1987:
what's the status on that one?

dirk-thomas:
I still think a warning should be printed and the behavior should be consistent with rmw_connext.

dhood:
I ran into this on OS X recently.

This is my understanding of the current state:
`connext_cmake_module`'s env hook:
- always sets the `NDDSHOME` to what was found at build time
- warns users if it doesn't match what is in the `NDDSHOME` env var

`opensplice_cmake_module`'s env hook:
- always sets the `OSPL_HOME` to what was found at build time
- does not have a warning if this doesn't match `OSPL_HOME` env var

---

And this is the direction we should go with this:
Both env hooks should:
- use the env var if it's set instead of forcing the one found at build time (to support users using fat archives)
- warn if this env var doesn't match what was found at build time.
    - this will always print a warning for users using fat archives (annoying)
    - will alert users that built from source to times when they are accidentally using a different version than that with which they built.

I'll work on consolidating these PRs

dirk-thomas:
> this will always print a warning for users using fat archives (annoying)

I don't think this is a good UX. If the build time path is still valid I would suggest to use that (without a warning).

dhood:
no, it's not. but this in the case where users have swapped to an abi incompatible version (the reason that you suggested the warning be added) then both paths can be valid. it's a bit late for feedback on this so we can either remove the warning completely for that section or live with it being printed for binary users. https://github.com/ros2/rmw_opensplice/pull/219 has the patch

dhood:
yeah, it's not really, that's what prompted me to point it out. However in the case where users have swapped to an abi incompatible version (the reason that you suggested the warning be added) then both paths can be valid; it's hard to distinguish between the two. #219 has the proposed patch, but it's a bit late for feedback on this so we can either remove the warning completely for that case or live with it being printed for binary users (I don't have time to test any additional logic).

dhood:
Updating with what I've understood to be the new plan for this.

Both env hooks should:
- use the build-time path if it is still valid.
- if the build-time path is not valid, use the env var if it's set.
- warn if no valid path was found (not error because perhaps we shipped a dual-homed fat archive and the user only wants to use one of the implementations)

As I understand it, a change of opinion has happened in this thread. We have gone from [preferring to use the env var](https://github.com/ros2/rmw_opensplice/pull/150#issuecomment-246715421) in the case that both are valid, to instead [always favouring the build-time path](https://github.com/ros2/rmw_opensplice/pull/150#issuecomment-350153981), provided it's valid. This seems reasonable to me, given that the build-time path is what it was actually built against; the env var support is mostly there for binary users. I am being explicit in case I misinterpreted something.

dirk-thomas:
Sounds good to me.

dhood:
absorbed into https://github.com/ros2/rmw_opensplice/pull/219

PR #154, opened 2016-10-30
match API change in rmw

mikaelarguedas:
follow-up of https://github.com/ros2/rmw/commit/baf307f343a36b6636894cfcd1175d54ed7a8798
This will throw an unused warning rather than an error next time we build rmw_opensplice_cpp

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_client.cpp
+++ b/rmw_opensplice_cpp/src/rmw_client.cpp
@@ -181,7 +181,7 @@ fail:
 }
 
 rmw_ret_t
-rmw_destroy_client(rmw_client_t * client)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Should this be updated to avoid the compiler warning too?

mikaelarguedas:
It should indeed. How do we plan on keeping track of the features we don't add to the rmw_implementations that are in "paused development". Add a TODO? ticket it? or both?

dirk-thomas:
I think it should definitely be a ticket since that is easy to discover when going to that repo. I don't mind about the TODO.

mikaelarguedas:
warning fixed.
Missing feature ticketed here: https://github.com/ros2/rmw_opensplice/issues/155

PR #157, opened 2016-11-17
expose template declaration publically and unify naming

dirk-thomas:
This will unify the symbols across the type support packages so that they only differ in the namespace.

Connect to ros2/rosidl_typesupport#1.

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_client.cpp
+++ b/rmw_opensplice_cpp/src/rmw_client.cpp
@@ -55,7 +55,7 @@ rmw_create_client(
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     type support,
     type_support->typesupport_identifier,
-    rosidl_typesupport_opensplice_cpp::typesupport_opensplice_identifier,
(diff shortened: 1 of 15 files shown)


rosidl_typesupport_opensplice_cpp/src/identifier.cpp
```diff
@@ -18,6 +18,6 @@ namespace rosidl_typesupport_opensplice_cpp
 {
 
 ROSIDL_TYPESUPPORT_OPENSPLICE_CPP_EXPORT
-const char * typesupport_opensplice_identifier = "opensplice_static";
+const char * typesupport_identifier = "opensplice_static";
```
mikaelarguedas: should this be `opensplice_static_cpp`, like for connext [here](https://github.com/ros2/rmw_connext/pull/206/files#diff-e5c49ab610395314d92f1faa255524bcR21) ?

dirk-thomas: The PR doesn't change the value of the identifier. It only aims to unify the naming of the symbols.

PR #161, opened 2017-01-12
update typesupport

dirk-thomas:
This patch updates the OpenSplice specific packages to the latest typesupport in order to be compilable again.

OpenSplice alone:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2238/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1695/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2215/)
  * Building OpenSplice alone doesn't work on Windows since in order to link the typesupport it needs to export the transitive libraries correctly: see `fatal error LNK1181: cannot open input file 'cmxml.lib'`. Since we don't support OpenSplice at the moment I won't spend any time on this.

Together with FastRTPS:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2237/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1698/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2224/)

It doesn't address any missing features and failing test since we don't actively support `rmw_opensplice_cpp` at the moment.


dirk-thomas:
Waiting for review.

rosidl_typesupport_opensplice_c/resource/srv__type_support_c.cpp.em
```diff
@@ -30,6 +30,7 @@
 #include "@(spec.pkg_name)/srv/dds_opensplice/ccpp_Sample_@(spec.srv_name)_Response_.h"
 #include "rosidl_typesupport_opensplice_c/identifier.h"
 
+// TODO(dirk-thomas) including another cpp file is just nasty
```
Karsten1987: Do you want to keep that comment?

dirk-thomas: Yes, because the following include of a cpp file is really bad practice and should be fixed.

dirk-thomas:
> I'd need a bit of explanation on that one. Does this mean we have generic type support and one specifically designed for OpenSplice?

Yes, since we use the generated message specific code for OpenSplice it needs its own typesupport to map from ROS types to the OpenSplice DDS types.

dirk-thomas:
> I'd need a bit of explanation on that one. Does this mean we have generic type support and one specifically designed for OpenSplice?

Yes, since we use the generated message specific code for OpenSplice it needs its own typesupport to map from ROS types to the OpenSplice DDS types.

There is no "generic typesupport". There is introspection typesupport but that only work when the rmw implementation supports something like the `DynamicData` API (which we use in `rmw_connext_dynamic_cpp` as well as `rmw_fastrtps_cpp`).

dirk-thomas:
Waiting for review.

Karsten1987:
as far as I can tell, this looks good to me.

PR #163, opened 2017-03-02
Use -Wpedantic

mikaelarguedas:
connects to https://github.com/ros2/ros2/issues/305

### DIFF
--- a/rmw_opensplice_cpp/src/typesupport_macros.hpp
+++ b/rmw_opensplice_cpp/src/typesupport_macros.hpp
@@ -30,9 +30,9 @@
       "type support handle implementation '%s' (%p) does not match valid type support " \
       "('%s' (%p))", \
       TYPE_SUPPORTS->typesupport_identifier, \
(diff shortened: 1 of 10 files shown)


rmw_opensplice_cpp/src/typesupport_macros.hpp
```diff
@@ -30,9 +30,9 @@
       "type support handle implementation '%s' (%p) does not match valid type support " \
       "('%s' (%p))", \
       TYPE_SUPPORTS->typesupport_identifier, \
-      TYPE_SUPPORTS->typesupport_identifier, \
+      reinterpret_cast<const void *>(TYPE_SUPPORTS->typesupport_identifier), \
```
dirk-thomas: Isn't a `static_cast` sufficient here?

mikaelarguedas: good point: https://github.com/ros2/rmw_opensplice/pull/163/commits/c91469e94864bd8c3cdb93377b5f8e73495b7427

PR #164, opened 2017-03-30
update to support new rmw API with namespaces

wjwwood:
Connects to ros2/rmw#95

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_node.cpp
+++ b/rmw_opensplice_cpp/src/rmw_node.cpp
@@ -28,10 +28,14 @@
 extern "C"
 {
 rmw_node_t *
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -202,6 +206,13 @@ rmw_create_node(const char * name, size_t domain_id)
   }
   memcpy(const_cast<char *>(node->name), name, strlen(name) + 1);
 
+  node->namespace_ = reinterpret_cast<const char *>(rmw_allocate(sizeof(char) * strlen(namespace_) + 1));
```
dirk-thomas: This memory is never freed. See #194.

PR #168, opened 2017-04-18
Add security_options to rmw_create_node argument list

mikaelarguedas:
connects to ros2/ros2#339

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_node.cpp
+++ b/rmw_opensplice_cpp/src/rmw_node.cpp
@@ -28,6 +28,17 @@
 extern "C"
 {
 rmw_node_t *
+rmw_create_secure_node(
+  const char * name,
+  const char * namespace_,
+  size_t domain_id,
+  char ** security_files_paths)
+{
+  (void)security_files_paths;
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Shouldn't this return an error code instead?

mikaelarguedas:
You're absolutely right, I'll return a runtime_error("Not implemented") here.

dirk-thomas:
This is a C API so the caller can't catch it in C. Therefore I think returning an error codec is more appropriate.

mikaelarguedas:
Yeah good point, returning an error code in a function that usually return a node structure is not convenient though. I'll return NULL and set an RMW error message then. Thanks for pointing it out

rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -38,6 +40,10 @@ rmw_create_node(const char * name, const char * namespace_, size_t domain_id)
     RMW_SET_ERROR_MSG("namespace_ is null");
     return nullptr;
   }
+  if (security_options->enforce_security) {
```
dirk-thomas: You must check that `security_options` is not null before accessing its members.

PR #162, opened 2017-01-26
expose node names

Karsten1987:
Connect to ros2/cli_tools#1

PR #170, opened 2017-04-20
fix cpplint failures

mikaelarguedas:
this fixes the style test failures introduced in https://github.com/ros2/rmw_opensplice/pull/164 and https://github.com/ros2/rmw_opensplice/pull/162

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_node.cpp
+++ b/rmw_opensplice_cpp/src/rmw_node.cpp
@@ -206,7 +206,8 @@ rmw_create_node(const char * name, const char * namespace_, size_t domain_id)
   }
   memcpy(const_cast<char *>(node->name), name, strlen(name) + 1);
 
-  node->namespace_ = reinterpret_cast<const char *>(rmw_allocate(sizeof(char) * strlen(namespace_) + 1));
+  node->namespace_ = reinterpret_cast<const char *>(
+    rmw_allocate(sizeof(char) * strlen(namespace_) + 1));
   if (!node->namespace_) {
     RMW_SET_ERROR_MSG("failed to allocate memory for node namespace");
     goto fail;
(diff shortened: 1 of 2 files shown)


rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -206,7 +206,8 @@ rmw_create_node(const char * name, const char * namespace_, size_t domain_id)
   }
   memcpy(const_cast<char *>(node->name), name, strlen(name) + 1);
 
-  node->namespace_ = reinterpret_cast<const char *>(rmw_allocate(sizeof(char) * strlen(namespace_) + 1));
+  node->namespace_ = reinterpret_cast<const char *>(
```
dirk-thomas: A `static_cast` should be sufficient here.

rmw_opensplice_cpp/src/rmw_node_names.cpp
```diff
@@ -31,5 +31,4 @@ rmw_get_node_names(
   RMW_SET_ERROR_MSG("get_node_names is not supported for Opensplice");
   return RMW_RET_ERROR;
 }
-
```
dirk-thomas: Why remove this empty line?

mikaelarguedas: because cpplint raises "/home/mikael/work/ros2/ros2_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_node_names.cpp:34:  Redundant blank line at the end of a code block should be deleted.  [whitespace/blank_line] [3]" otherwise

rmw_opensplice_cpp/src/rmw_node.cpp
mikaelarguedas: the same casting logic has been used in all the namespace implementation PRs (e.g. https://github.com/ros2/rmw_connext/pull/220/files#diff-9bba412d4597100825da2b2307403d08R436). So I'm not going to change it in this one. Feel free to open a set of follow-up PRs to fix the cast logic all at once

PR #172, opened 2017-06-07
update to use OpenSplice 6.7

dirk-thomas:
This is the first step getting the ROS 2 code to compile against OpenSplice 6.7.

The implementation needs to be updated with all the recent changes (partitions, topic prefixes, etc.). So currently many tests are still failing. Also CI is not able to use OpenSplice 6.7 atm.

### DIFF
--- a/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
+++ b/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
@@ -44,7 +44,10 @@
 set(OpenSplice_FOUND FALSE)
 
 # check if provided OSPL_HOME is from an "official" binary package
-file(TO_CMAKE_PATH "$ENV{OSPL_HOME}" _ospl_home)
+set(_ospl_home "")
+if(DEFINED ENV{OSPL_HOME})
+  file(TO_CMAKE_PATH "$ENV{OSPL_HOME}" _ospl_home)
+endif()
 if(WIN32)
   set(_ospl_release_file release.bat)
 else()
--- a/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
+++ b/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
@@ -76,7 +76,6 @@ if(NOT _ospl_home STREQUAL "")
     "ddsserialization"
     "ddsuser"
     "ddsutil"
-    "durability"
     "spliced"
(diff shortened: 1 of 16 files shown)


dirk-thomas:
@mikaelarguedas I can address the deprecated option but likely in a separate PR. I would first like to get this into a state where master can be CI tested again.

This patch finally passes the build phase on Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2879/)

It uses a custom Debian which will be replaced with a new "final" .deb in the process of being built and placed in packages.osrfoundation.org. Once that is available I will make a corresponding PR to the CI repo.

Therefore this PR is ready for review.

opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
```diff
@@ -76,7 +79,7 @@ if(NOT _ospl_home STREQUAL "")
     "ddsserialization"
     "ddsuser"
     "ddsutil"
-    "durability"
+    # "durability"
```
mikaelarguedas: does that library not exist at all ? or just not built properly in our custom deb ?
I think we should either put a comment here or remove the line depending on what makes most sense

rosidl_typesupport_opensplice_cpp/resource/msg__type_support.cpp.em
```diff
@@ -28,6 +28,7 @@
 #include "rosidl_typesupport_opensplice_cpp/identifier.hpp"
 #include "rosidl_typesupport_opensplice_cpp/message_type_support.h"
 #include "rosidl_typesupport_opensplice_cpp/message_type_support_decl.hpp"
+#include "rosidl_typesupport_opensplice_cpp/u__instanceHandle.h"
```
mikaelarguedas: is the semantic of this header different ? (to use the double underscore while every other header has a single one)

dirk-thomas: The file is coming from OpenSplice. I didn't change their naming scheme.

opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
dirk-thomas: Good point. As far as I know it doesn't exist anymore. I will update the commit which commented it out to actually remove the line.

PR #181, opened 2017-08-17
Implement avoid_ros_namespace_conventions

michielb-prismtech:
- added new function process_service_name
- remove datawriter&datareader qos from requester/responder signature and replace with rmw_qos_profile_t
- move qos conversion functions from rmw_opensplice to rosidl_typesupport_opensplice_cpp

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_client.cpp
+++ b/rmw_opensplice_cpp/src/rmw_client.cpp
@@ -58,11 +58,6 @@ rmw_create_client(
     return nullptr;
   }
 
-  if (qos_profile->avoid_ros_namespace_conventions) {
-    RMW_SET_ERROR_MSG("QoS 'avoid_ros_namespace_conventions' is not implemented");
-    return NULL;
-  }
-
   auto node_info = static_cast<OpenSpliceStaticNodeInfo *>(node->data);
   if (!node_info) {
     RMW_SET_ERROR_MSG("node info handle is null");
@@ -112,6 +107,7 @@ rmw_create_client(
     reinterpret_cast<void **>(&response_datareader),
     reinterpret_cast<const void *>(&datareader_qos),
     reinterpret_cast<const void *>(&datawriter_qos),
+    qos_profile->avoid_ros_namespace_conventions,
     &rmw_allocate);
   if (error_string) {
     RMW_SET_ERROR_MSG((std::string("failed to create requester: ") + error_string).c_str());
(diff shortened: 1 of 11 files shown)


mikaelarguedas:
I didn't review then code changes but launched CI for sanity checking.
The job [failed to build](http://ci.ros2.org/job/ci_linux/3031/) because `rmw_typesupport_opensplice_cpp` [find_package's rmw](https://github.com/ros2/rmw_opensplice/pull/181/files#diff-086687f92f26520a8cd384557cba7109R24) without declaring a dependency on it in its `package.xml`

dirk-thomas:
While this could be addressed by declaring a build dependency on `rmw` I am not sure why `qos.cpp` was moved to the type support. I think it would be "better" from a decoupling point of view if the type support package would not depend on the `rmw` interface. As far as I know non of the other type support packages depend on `rmw`.

@michielb-prismtech Can you explain why you moved the code from the `rmw_opensplice_cpp` package to the type support. Maybe there is a way which works without moving it (and introducing this dependency).

michielb-prismtech:
Before the topic_name/service_name change (the split between topic and partition names) the create_requester/create_responder required a datawriter_qos and datareader_qos arguments to be created. In the new situation they also require avoid_ros_namespace_conventions (part of the rmw_qos_profile_t) to determine the partition name. And since the datareader_qos, datawriter_qos and avoid_ros_namespace_conventions were all determined based on the rmw_qos_profile_t, I changed the signature of the create_requester and create_responder to take a rmw_qos_profile_t argument and removed the reader and writer qos. And because of the signature change the qos transformation functions where now required in the typesupport and in rmw_opensplice and they both have access to typesupport hence the move.
I think it should also be possible to change the signature for the create_requester/create_responder to add the avoid_ros_namespace_conventions boolean. This way the qos transformation can be moved back to rmw_opensplice. So if you rather have it that way I'm happy to change it.

dirk-thomas:
> I think it should also be possible to change the signature for the create_requester/create_responder to add the avoid_ros_namespace_conventions boolean. This way the qos transformation can be moved back to rmw_opensplice. So if you rather have it that way I'm happy to change it.

Thanks for the detailed description. That would be great. While I can't point to a specific reason the typesupport should not depend on rmw I think it is a better separation if it doesn't. That allows to build all message packages which need the typesupport independent of the rmw package.

dirk-thomas:
I triggered a new build: [Build Status](http://ci.ros2.org/job/ci_linux/3056/)

michielb-prismtech:
Is anything more expected from me on this issue?

dirk-thomas:
I think this can be merged as-is. I assume the remaining tests failures will be address by something else. Please feel free to merge. :+1:

michielb-prismtech:
Yes, we're adding a couple of fixes to OpenSplice CE soon.

PR #185, opened 2017-08-23
Use matched count to determine server_is_available

michielb-prismtech:
- change the signature of internal server_is_available, removed the count_xxx function pointer
- the server_is_available function now use publication_matched and subscription_matched functions
- DCPSPublication and DCPSSubscription listeners add partition name for discovered topics

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
+++ b/rmw_opensplice_cpp/src/rmw_service_server_is_available.cpp
@@ -62,8 +62,7 @@ rmw_service_server_is_available(
     return RMW_RET_ERROR;
   }
 
-  const char * err = callbacks->server_is_available(
-    client_info->requester_, node, is_available, &rmw_count_publishers, &rmw_count_subscribers);
+  const char * err = callbacks->server_is_available(client_info->requester_, node, is_available);
   if (err) {
     RMW_SET_ERROR_MSG(err);
     return RMW_RET_ERROR;
--- a/rmw_opensplice_cpp/src/types.cpp
+++ b/rmw_opensplice_cpp/src/types.cpp
@@ -19,7 +19,9 @@
 #include <map>
 #include <set>
 #include <string>
+#include <vector>
 
+#include "rosidl_typesupport_opensplice_cpp/misc.hpp"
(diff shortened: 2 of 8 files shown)


dirk-thomas:
LGTM. Do you want me to trigger a CI build for this? Otherwise please feel free to merge this.

michielb-prismtech:
Yes I like a CI build, tested locally and build fine just like to make sure I don't push anything broken into master

dirk-thomas:
[Build Status](http://ci.ros2.org/job/ci_linux/3068/)

michielb-prismtech:
I agree, the current listener implementation adds the topic_names (incl partition) to a list and the count functions return the count of publishers/subscribers based on the topic_name match with the list. This check is incomplete, QoS also needs to be compatible in order to count matches. 
Following functions are used to get the count:
```
rmw_ret_t rmw_count_subscribers(const rmw_node_t * node, const char * topic_name, size_t * count);
rmw_ret_t rmw_count_publishers(const rmw_node_t * node,  const char * topic_name, size_t * count);
```
I suggest adding new functions (and maybe even remove the previous count functions)
```
rmw_ret_t rmw_subscriber_count(const rmw_subscriber_t *, size_t *);
rmw_ret_t rmw_publisher_count(const rmw_publisher_t *, size_t *);
```
with these function you could then get the actual number of matching publishers/subscribers. Extra information available in the get_publication_matched_status() could also be added.

dirk-thomas:
It would indeed be good to add this to the `rmw` API. I created a ticket based on your earlier feedback: ros2/rmw#120.

I am not sure if you would like to contribute this addition to the `rmw` interface as well as expose these functions afterwards in `rcl*`?

michielb-prismtech:
I can, but first need to make sure that the beta3-release fixes for opensplice are ready.

michielb-prismtech:
Think I should have rebased instead

dirk-thomas:
> I can, but first need to make sure that the beta3-release fixes for opensplice are ready.

Absolutely. Before you look at the `rmw` ticket please comment there so other are aware of it to avoid duplicate efforts.

PR #191, opened 2017-08-31
Implement rmw_get_node_names

michielb-prismtech:
setting node names not natively supported in OpenSplice (or DDS spec). Added the node name
to the userData in the ParticipantQos in such a way that userData can still be used for other
extensions in the future.
Rest of implementation based on rmw_connext

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_node.cpp
+++ b/rmw_opensplice_cpp/src/rmw_node.cpp
@@ -118,8 +118,27 @@ rmw_create_node(
   }
 #endif
 
+  // Communicate the ROS Node name as DDS Builtin Participant Topic userData
+  //   For extensibility and backwards compatibility a format is applied that must be understood.
+  //   The format specifies a list of name-value pairs,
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Can you clarify which part of currently used node names are not supported by the DDS spec.

michielb-prismtech:
There is no mechanism specified to distribute the node name. 

RTI has added proprietary extensions to the built-in DCPSParticipant topic, one of which is a participant_name. They're using the participant_name which is set on the DomainParticipant via the DomainParticipantQos. In the rmw_create_node the participant is created and the name is set and DCPSParticipant topic data is published . On remote nodes the built-in DCPSParticipant topic data is received and is accessible via the rmw_get_node_names function.
According to the spec the DCPSParticipant only has two fields a **key** and **user_data**.

In you're implementation you want the node names, which is a user requirement so we added it to the user_data with this commit.

dirk-thomas:
Thank you for the detailed explanation. New CI build with this branch: [Build Status](http://ci.ros2.org/job/ci_linux/3122/)

PR #192, opened 2017-08-31
fix wrong expectation about output srv files

### LINKED ISSUE #171: recompliation of messages always happens
Whenever my message projects are recompiled (using Ament), regardless of whether any code has changed or not, the rmw_opensplice message support gets rebuilt. This is painful and expensive, costing me an hour of compilation time every day. Is there something I can do to make middleware support be smart enough to not recompile message packages when nothing has changed?

dirk-thomas:
Fixes #171.


mikaelarguedas:
looks good to me. Does that fix https://github.com/ros2/rmw_opensplice/issues/171 or is it just a first step toward not having the duplicated message generation ?

dirk-thomas:
I think it does what it says :wink:

> Fixes #171.

mikaelarguedas:
oops just looked at the diff not the description :S

PR #196, opened 2017-09-08
Compiler warning fixes for Windows

michielb-prismtech:


rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -133,7 +133,7 @@ rmw_create_node(
   DDS::DomainParticipantQos dpqos;
   size_t length = strlen(name) + strlen("name=;") + 1;
   dp_factory->get_default_participant_qos(dpqos);
-  dpqos.user_data.value.length(length);
+  dpqos.user_data.value.length((DDS::ULong)length);
```
dirk-thomas: Please avoid C casts in C++ code and use `static_cast<DDS::ULong>(length)` instead.

rosidl_typesupport_opensplice_c/resource/msg__type_support_c.cpp.em
```diff
@@ -397,7 +397,7 @@ else:
       return "failed to assign string into field '@(field.name)'";
     }
 @[  elif field.type.is_primitive_type()]@
-    ros_message->@(field.name) = dds_message->@(field.name)_@(' == TRUE' if field.type.type == 'bool' else '');
+    ros_message->@(field.name) = @('((' if field.type.type == 'bool' else '')dds_message->@(field.name)_@(')? true : false)' if field.type.type == 'bool' else '');
```
dirk-thomas: I haven't tried to run this line yet but it looks "fishy" :wink: Can you describe what it is supposed to do since I can't decipher the template?

michielb-prismtech: Have to get back to you on that one, as I'm not the one who made it

michielb-prismtech: btw cannot be run on windows without #197

dirk-thomas: I will trigger a new build once #197 passed CI and has been merged.

dirk-thomas:
* Windows (non-parallel) [Build Status](http://ci.ros2.org/job/ci_windows_opensplice67/9/)

dirk-thomas:
* Windows (non-parallel) [Build Status](http://ci.ros2.org/job/ci_windows_opensplice67/10/)

dirk-thomas:
The latest build failed without using `--parallel` with the following error:

```
cppgen : error : unable to open file: JointState_.h
```

This seems to be like a bug in `cppgen`. We have seen it before on other builds. @michielb-prismtech Can you please try to debug this locally (on Windows) and see if you can reproduce it?

dirk-thomas:
* Windows (non-parallel) [Build Status](http://ci.ros2.org/job/ci_windows_opensplice67/11/)

dirk-thomas:
Hm, the next build failed with:

```
PermissionError: [Errno 13] Permission denied: 'C:/J/workspace/ci_windows_opensplice67/ws/build/pendulum_msgs/rosidl_typesupport_opensplice_cpp/pendulum_msgs\\msg\\dds_opensplice\\JointState_Dcps_impl.h'
```
:thinking: :question:

dirk-thomas:
Often this is due to some application like "Windows Defender" holding a file handle and therefore letting the access fail. But the Jenkins folder is explicitly listed as an excluded path on our Windows nodes (just double checked).

dirk-thomas:
After discussing this offline with @mikaelarguedas we think this is the same bug we hit with CMake on Windows in ros2/rosidl#124.

The `add_library` target in the C++ typesupport is depending on the `add_custom_command` in the C++ typesupport. Additionally the `add_custom_command` in the C typesupport is depending in `add_custom_command` in the C++ typesupport. Since the custom command isn't defined in a separate file (as mentioned in the referenced ticket) this will trigger the custom command of the C++ typesupport twice :bomb:

Instead of separating the custom command into its own file (as in the referenced PR) we removed the direct dependency between the two custom commands in 0693b8c56ee02f6bdddfcf486c72b4a2f0594e48. This is possible because the C typesupport generation script doesn't actually need the source files generated by the C++ typesupport yet. It can expand its templates just fine without them being there. Since the library target in the C typesupport depends on the library target in the C++ typesupport (and not on the custom command of the C++ typesupport) that should work just fine :crossed_fingers:

[Build Status](http://ci.ros2.org/job/ci_windows_opensplice67/13/)

dirk-thomas:
I squashed all but the last commit.

mikaelarguedas:
glad we figured it out, this code is not the most trivial to investigate... and the result is a cleaner dependency tree :clap: :tada:

PR #200, opened 2017-09-12
Initial rmw_get_service_names_and_types

michielb-prismtech:
Just an initial implementation not yet tested.

### DIFF
--- a/rmw_opensplice_cpp/CMakeLists.txt
+++ b/rmw_opensplice_cpp/CMakeLists.txt
@@ -46,6 +46,7 @@ link_directories(${OpenSplice_LIBRARY_DIRS})
 add_library(rmw_opensplice_cpp SHARED
   src/demangle.cpp
   src/identifier.cpp
+  src/namespace_prefix.cpp
   src/qos.cpp
   src/rmw_client.cpp
   src/rmw_compare_gids_equal.cpp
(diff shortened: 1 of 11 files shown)


dirk-thomas:
We haven't tagged the Beta 3 release yet but are planning to do so today. What is the state of this? Have you had a chance to test this?

michielb-prismtech:
I've found that the implementation did not work, I've fixed it. However I've not been able to run the complete testsuite yet. My hope is that the CI test run doesn't show any failures.

dirk-thomas:
@michielb-prismtech This PR currently lacks the new file `src/demangle.cpp`. Once you have added that I can trigger new CI builds.

michielb-prismtech:
typed my message before I pushed, will do it other way around next time.

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3230/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3292/)

If the CI builds pass green are you confident to merge this change for Beta 3 (since we are pretty late in the testing phase). Merging it has the (+) that it aims to support the service introspection. On the (-) side there is a chance of regression with any patch (we won't have time to run another round of testing before cutting the release today). I am ok with merging it if you think that is the right thing.

michielb-prismtech:
I wait for the results before answering that question ;)

michielb-prismtech:
Test results look good, if you agree with the changes I'm confident the fix is good. Moved some functions to new files copied two from connext and implemented a get_service_names_and_types function which did not function before. Greatest risk is the newly implemented function but as it didn't function before it's not a big risk

dirk-thomas:
Sounds good to me: :ship:

Thank you for your effort on this!

PR #201, opened 2017-09-13
correct responder/requester topic names

michielb-prismtech:
During running the checklist we found that the topic names for the requester and responder did not match with fastRTPS implementation

### DIFF
--- a/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
+++ b/rosidl_typesupport_opensplice_cpp/include/rosidl_typesupport_opensplice_cpp/requester.hpp
@@ -107,8 +107,8 @@ public:
       estr = "process_service_name: failed";
       goto fail;
     }
-    request_topic_name = service_str + "_Request";
-    response_topic_name = service_str + "_Response";
+    request_topic_name = service_str + "Request";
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
@michielb-prismtech It's not obvious to me that there is anything wrong with the current implementation (while it is indeed different from Fast-RTPS's). Looking at the [design doc](http://design.ros2.org/articles/topic_and_service_names.html) it's not clear to me how the request/response topic should be named other than having the corresponding `rq` or `rr` prefix.
Maybe @wjwwood can give us more details about what the expected name for these topics should be

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3227/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3287/)

michielb-prismtech:
How is this trival change failing on windows?

dirk-thomas:
I guess the previous Windows job failed because someone was "working" on that Windows node at the same time: [Build Status](http://ci.ros2.org/job/ci_windows/3290/)

dirk-thomas:
@michielb-prismtech This is ready to be merged. How is it related to @200?

michielb-prismtech:
When testing #200 this was noticed. No dependency.

PR #203, opened 2017-09-14
Trigger publication/subscription_listener

michielb-prismtech:
To handle missed event, manually trigger listener callback

The publisher_listener and subscriber_listener are set on an already created readers. This could mean that the readers already contains data for which the on_data_available callback is not called. To ensure we have not missed anything manually call the on_data_available callback. Because of the locking in the on_data_available functions it is possible to call this from multiple threads.

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_node.cpp
+++ b/rmw_opensplice_cpp/src/rmw_node.cpp
@@ -208,6 +208,10 @@ rmw_create_node(
     publisher_listener, buf, goto fail, CustomPublisherListener, graph_guard_condition)
   buf = nullptr;
   builtin_publication_datareader->set_listener(publisher_listener, DDS::DATA_AVAILABLE_STATUS);
+  // listener is set on an already created reader. This could mean that the reader already contains
+  // data for which the on_data_available callback is not called. To ensure we have not missed
+  // anything manually call the on_data_available callback.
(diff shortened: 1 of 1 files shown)


dirk-thomas:
We have tagged Beta 3 yesterday. Can you describe what issues users might experience due to this so that it can be mentioned in the known issues list.

michielb-prismtech:
Not really sure how this impact the users but a rmw_node has a builtin readers for DCPSPublication and DCPSSubscription. The data form those topics is read by a reader on a on_data_available event. Because the listeners are set after the reader is created it is possible to have missed events (data is not lost). The rmw_get_service_names_and_types and rmw_get_topic_names_and_types function use the data DCPSPublication and DCPSSubscription data to determine what is available. So depending on when/if a new event is triggered it will take time before the data required for those functions is available. A new event is generated on creation and deletion of readers and writers (also readers/writers within node) so I most cases I think an event will be generated. Don't know if other functions are dependent on this mechanism.

dirk-thomas:
Thanks for clarifying. Sounds like a good improvement to avoid races but nothing severe beta 3 users will need to worry about.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3232/)
* Linux aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/540/)
* Wwindows [Build Status](http://ci.ros2.org/job/ci_windows/3295/)

PR #207, opened 2017-10-23
Fix rmw_opensplice dead code

gaoethan:
Dead code in API `rmw_create_subscription()`
the condition "`subscription`" cannot be true in the "`fail`"
clauses and the execution never reaches this statement:

```
if (subscription) {
    rmw_subscription_free(subscription);
  }
```

the "`subscribtion`" will be finally free by
`rmw_destroy_subscription() `when it's not a nullptr

Dead code in API `rmw_create_publisher()`
the condition "`publisher`" cannot be true in the "`fail`"
clauses, so the execution cannot reach this statement:

```
if (publisher) {
    rmw_publisher_free(publisher);
  }
```

the "`publisher`" will be finally free by
`rmw_destroy_publisher()` when it's not a nullptr

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_publisher.cpp
+++ b/rmw_opensplice_cpp/src/rmw_publisher.cpp
@@ -193,9 +193,6 @@ rmw_create_publisher(
 
   return publisher;
 fail:
-  if (publisher) {
-    rmw_publisher_free(publisher);
-  }
   if (dds_publisher) {
     if (topic_writer) {
       status = dds_publisher->delete_datawriter(topic_writer);
--- a/rmw_opensplice_cpp/src/rmw_subscription.cpp
+++ b/rmw_opensplice_cpp/src/rmw_subscription.cpp
@@ -226,9 +226,7 @@ fail:
   if (subscriber_info) {
     rmw_free(subscriber_info);
   }
-  if (subscription) {
-    rmw_subscription_free(subscription);
-  }
+
   if (buf) {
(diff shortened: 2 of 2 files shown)


rmw_opensplice_cpp/src/rmw_publisher.cpp
```diff
@@ -193,9 +193,6 @@ rmw_create_publisher(
 
   return publisher;
 fail:
-  if (publisher) {
```
dirk-thomas: Wouldn't this be the case when the goto from the following line triggers: https://github.com/ros2/rmw_opensplice/pull/207/files#diff-2184ec33b792d378e335481d04c19f40R190

rmw_opensplice_cpp/src/rmw_subscription.cpp
```diff
@@ -226,9 +226,7 @@ rmw_create_subscription(
   if (subscriber_info) {
     rmw_free(subscriber_info);
   }
-  if (subscription) {
-    rmw_subscription_free(subscription);
-  }
```
dirk-thomas: Wouldn't this be the case when the goto from the following line triggers: https://github.com/ros2/rmw_opensplice/pull/207/files#diff-153e779bdc552cb4e36e9029830b2307R197

gaoethan:
@dirk-thomas thanks for your catching.

PR #209, opened 2017-10-29
API to manipulate log setting macros

sriramster:
Enable manipulating logging macros from rmw_opensplice.

Connect to ros2/rmw#124


rmw_opensplice_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "rmw/rmw.h"
+
+extern "C"
+{
+// TODO(sriramster): Check support for log level manipulation with opensplice
+rmw_ret_t
+rmw_set_log_severity(rmw_log_severity_t severity)
+{
+  return RMW_RET_OK;
```
dirk-thomas: As long as this is not implemented it shouldn't return "ok".

dirk-thomas: Can you look into implementing this function or do propose to merge this as-is?

sriramster:
@dirk-thomas @dhood Could you please review this patch? This is connected to another approved patch by Dirk

dirk-thomas:
> Could you please review this patch?

I am still waiting for an answer on my [previous question](https://github.com/ros2/rmw_opensplice/pull/209#discussion_r149206143).

rmw_opensplice_cpp/src/rmw_logging.cpp
sriramster: @dirk-thomas Please merge as-is. I'll try to push in these changes after working on couple more patches.

PR #212, opened 2017-11-11
avoid repeated finding

dirk-thomas:
Similar to ros2/rcl#177.

### DIFF
--- a/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
+++ b/opensplice_cmake_module/cmake/Modules/FindOpenSplice.cmake
@@ -41,6 +41,10 @@
 
 # lint_cmake: -convention/filename, -package/stdargs
 
+if(DEFINED OpenSplice_FOUND)
+  return()
+endif()
+
 set(OpenSplice_FOUND FALSE)
 
 # check if provided OSPL_HOME is from an "official" binary package


dirk-thomas:
Waiting for review.

dhood:
Considering users who built without opensplice and then install opensplice and want to re-build with opensplice support: is `--force-cmake-configure` enough for them to "reset" this check?

dirk-thomas:
Yes, since the variable is not stored in the cache it will be re-evaluated at CMake configure time.

dhood:
Thinking about other side-effects: This could change the experience for users: previously they would have seen instantaneous impact of modifying e.g. `OSPL_HOME` between re-builds. I think requiring them to use `--force-cmake-configure` now for that case is reasonable, and I don't see the change in behaviour significant enough to warrant documentation.

There's also a chance that users will interrupt the first finding and then it will require `--force-cmake-configure` to re-attempt it. This can be alleviated if we move setting `OpenSplice_FOUND` to `FALSE` further down until the conclusion is made. Mostly for my learning: do you think this is worthwhile, or it's not typically our job to support interruptions gracefully? (you might have already decided against it, or might not have considered it).

dirk-thomas:
This patch does not affect either of the cases you describe. Before when CMake was run the same logic was evaluates multiple times (always coming to the same conclusion). After the patch the logic is only being executed once which defines the variable and on consecutive invocations it is not being re-computed.

E.g. when you interrupt CMake and a package didn't finish to (re-)configure it will do so again on the next invocation again.

dhood:
Ok, I can see that it will not matter if the user interrupts the build because the package won't have been fully configured and therefore will retry, thank you.

And for the situation where a user modifies `OSPL_HOME`, they would have had to use `--force-cmake-configure` even before to get the package to re-run the configuration step in order to use the new settings.

PR #217, opened 2017-12-05
remove special handling of builtin_interfaces

### LINKED ISSUE #216: publishing or subscribing with builtin_interfaces Time (and likely Duration) fails with an assertion
rmw_opensplice special cases builtin_interfaces, like [here](https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_c/cmake/rosidl_typesupport_opensplice_c_generate_interfaces.cmake#L24)

However this is causing failures as observed testing ros2/rclcpp#407

The underlying problem is that get_message_type_support_handle is returning null for builtin_interfaces datatypes. 
```
  auto time_typesupport_handle = rosidl_typesupport_cpp::get_message_type_support_handle<builtin_interfaces::msg::Time>();
  std::cout << "Time typesupport handle is " << time_typesupport_handle << std::endl;

  auto parameter_typesupport_handle = rosidl_typesupport_cpp::get_message_type_support_handle<rcl_interfaces::msg::Parameter>();
  std::cout << "Parameter typesupport handle is " << parameter_typesupport_handle << std::endl;
```
Result:
```
Time typesupport handle is 0
Parameter typesupport handle is 0x7ffff60912e0
```

This method is used directly in the rclcpp [subscription constructor](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/subscription.hpp#L160).
Which causes the constructor to fail.

This also is an issue for publishing [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/publisher.hpp#L176)

dirk-thomas:
Fixes #216. This is one way to address the issue.


dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3741/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/893/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3082/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3837/)

dirk-thomas:
Waiting for review...

mikaelarguedas:
> The only potential issue looks to be test_composition failing on OSX with fastrtps

Yeah thats a test that got flakier recently it's failing quite a lot on all platforms with fastrtps so I think it's safe to consider the failure not related to this PR

PR #232, opened 2018-06-20
homogenize opensplice keys and use libopensplice67 everywhere

mikaelarguedas:


mikaelarguedas:
for additional context: I decided to call the version number explcitly in case we release another version of this package during bionic's lifetime.

I'll leave a change to @nuclearsandwich to give feedback before merging because handling the work of managing DDS impl dependencies in the various packages at the moment and may conflict with ongoing work on nuclearsandwich's side.

nuclearsandwich:
> I'll leave a change to @nuclearsandwich to give feedback before merging because handling the work of managing DDS impl dependencies in the various packages at the moment and may conflict with ongoing work on nuclearsandwich's side.

I don't have anything in flight for opensplice that would interfere with this. Using consistent keys is the big thing :+1:

PR #234, opened 2018-06-22
strip 'Sample_' from reported types

mikaelarguedas:
Opening this for visibility.
I didn't see anywhere the `Sample_` prefix was defined so I ended up hard-coding it in here for now.

Feedback on the approach is welcome.

Connects to ros2/ros2cli#106

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp
+++ b/rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp
@@ -30,6 +30,7 @@
 #include "types.hpp"
 #include "demangle.hpp"
 
+#define SAMPLE_PREFIX "/Sample_"
 // The extern "C" here enforces that overloading is not used.
 extern "C"
 {
@@ -119,7 +120,13 @@ rmw_get_service_names_and_types(
       // Duplicate and store each type for the service
       size_t type_index = 0;
       for (const auto & type : service_n_types.second) {
-        char * type_name = rcutils_strdup(type.c_str(), *allocator);
+        size_t n = type.find(SAMPLE_PREFIX);
+        std::string stripped_type = type;
+        if (n < type.size() - strlen(SAMPLE_PREFIX)) {
+          stripped_type = std::string(type.substr(0, n + 1).c_str()) + \
+            std::string(type.substr(n + strlen(SAMPLE_PREFIX)).c_str());
+        }
+        char * type_name = rcutils_strdup(stripped_type.c_str(), *allocator);
         if (!type_name) {
           RMW_SET_ERROR_MSG_ALLOC("failed to allocate memory for type name", *allocator)
           fail_cleanup();


rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp
```diff
@@ -119,7 +120,13 @@ rmw_get_service_names_and_types(
       // Duplicate and store each type for the service
       size_t type_index = 0;
       for (const auto & type : service_n_types.second) {
-        char * type_name = rcutils_strdup(type.c_str(), *allocator);
+        size_t n = type.find(SAMPLE_PREFIX);
+        std::string stripped_type = type;
+        if (n < type.size() - strlen(SAMPLE_PREFIX)) {
```
dirk-thomas: I would change the condition to `n != std::string::npos`.

rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp
```diff
@@ -119,7 +120,13 @@ rmw_get_service_names_and_types(
...
+          stripped_type = std::string(type.substr(0, n + 1).c_str()) + \
```
dirk-thomas: `type.substr()` already returns a `std::string` so no need to std::string(result.c_str())` the result.

Same on the next line.

mikaelarguedas:
> The question is: is it guaranteed that the service name contains Sample_? If yes, I think the conditional check should be removed and it should ensured that it is being found. If no, then the approach is fragile since a type which has the same prefix but is intentionally named like that by the user would be mangled.

Yes I agree and share this concern. In the case of ROS services I think it will always contain that prefix.
For native ones it could be anything, but it's not really something we support ATM.

> If no, then the approach is fragile since a type which has the same prefix but is intentionally named like that by the user would be mangled.

Agreed, I didnt know how to make it less fragile at a glance that's why I used that approach.
I'll remove the check for now then.
As we don't allow slashes in package names of message names, it should be safe enough to use `/Sample_` in the ROS case

dirk-thomas:
I inverted the logic in d45c98b5ed11a42dc5489ff0b1a580694482be03 to match the other error handling in this function. It gets rid of the `else` as well as the early defintition of `stripped_type`.

I did the edit in the GitHub UI so I didn't build / lint the code - but looks fine to me :wink:

rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp
```diff
@@ -119,7 +120,17 @@ rmw_get_service_names_and_types(
       // Duplicate and store each type for the service
       size_t type_index = 0;
       for (const auto & type : service_n_types.second) {
-        char * type_name = rcutils_strdup(type.c_str(), *allocator);
+        size_t n = type.find(SAMPLE_PREFIX);
+        if (std::string::npos == n) {
+          RMW_SET_ERROR_MSG_ALLOC(
+            "failed to convert DDS type name to ROS service type name: '"
+            SAMPLE_PREFIX "' not found",
```
dirk-thomas: Sorry, another late note: maybe printing the actual `type` here would make the error message more useful.

mikaelarguedas:
error message with https://github.com/ros2/rmw_opensplice/pull/234/commits/b4b6cef170712bd6204c6bc8d6d87b4685bef956
```
Failed to get_service_names_and_types: failed to convert DDS type name to ROS service type name: '/Sample_' not found in type: 'myROS/ServiceType', at /home/mikael/work/ros2/bouncy_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp:129
```

rmw_opensplice_cpp/src/rmw_service_names_and_types.cpp
```diff
@@ -119,7 +120,18 @@ rmw_get_service_names_and_types(
       // Duplicate and store each type for the service
       size_t type_index = 0;
       for (const auto & type : service_n_types.second) {
-        char * type_name = rcutils_strdup(type.c_str(), *allocator);
+        size_t n = type.find(SAMPLE_PREFIX);
+        if (std::string::npos == n) {
+          char * error_msg = rcutils_format_string(
```
dirk-thomas: The returned char * needs to be deallocated in order to not leak.

tfoote:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4799/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1660/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3984/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4853/)

PR #241, opened 2018-07-16
add rmw_get_encoding_identifier

Karsten1987:
connects to ros2/rmw#143


rmw_opensplice_cpp/src/rmw_get_encoding_identifier.cpp
```diff
@@ -0,0 +1,27 @@
+// Copyright 2014-2015 Open Source Robotics Foundation, Inc.
```
wjwwood: Are the dates right? (might want to check the other related pr's I already approved too)

PR #246, opened 2018-11-20
refactor to support init options and context

wjwwood:
Connects to ros2/rmw#154

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_init.cpp
+++ b/rmw_opensplice_cpp/src/rmw_init.cpp
@@ -30,15 +30,77 @@
 #include <dds_dcps.h>
 
 #include "rmw/error_handling.h"
+#include "rmw/impl/cpp/macros.hpp"
 #include "rmw/rmw.h"
 #include "rmw/types.h"
+
+#include "identifier.hpp"
 
 // The extern "C" here enforces that overloading is not used.
 extern "C"
 {
 rmw_ret_t
-rmw_init()
+rmw_init_options_init(rmw_init_options_t * init_options, rcutils_allocator_t allocator)
 {
+  RMW_CHECK_ARGUMENT_FOR_NULL(init_options, RMW_RET_INVALID_ARGUMENT);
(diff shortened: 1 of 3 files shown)


rmw_opensplice_cpp/src/rmw_init.cpp
```diff
@@ -16,20 +16,96 @@
...
+  }
+  *dst = *src;
+  return RMW_RET_OK;
+}
+
+rmw_ret_t
+rmw_init_options_fini(rmw_init_options_t * init_options)
+{
+  RMW_CHECK_ARGUMENT_FOR_NULL(init_options, RMW_RET_INVALID_ARGUMENT);
+  RCUTILS_CHECK_ALLOCATOR(&(init_options->allocator), return RMW_RET_INVALID_ARGUMENT);
```
jacobperron: Do we care if the alllocator is invalid at this point? It doesn't appear to be used.

wjwwood: On the one hand, no because we're not using it, but on the other hand all other rmw implementations will expect it to be valid, so I think it's a good check to have.

rmw_opensplice_cpp/src/rmw_init.cpp
```diff
@@ -16,20 +16,96 @@
...
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    opensplice_cpp_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
```
sloretz: Looks like this has the same check for null

PR #250, opened 2018-11-27
suppress warning on Windows

dirk-thomas:
Follow up of #249.

Almost green Windows build including the external `cdr_serialization` branches: [Build Status](https://ci.ros2.org/job/ci_windows/5564/)


dirk-thomas:
Since the patch is matching what has been applied and approved for `clang` before I will go ahead and merge this. Please feel free to comment after the fact with any feedback.

PR #263, opened 2019-03-20
Use regex for checking namespace for ROS-specific types

jacobperron:
This enables support for ROS message types with namespaces other than 'msg' or 'srv' (e.g. 'action').

---

Related to ros2/ros2cli#202. This is needed for querying action types to support the Action CLI.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6483/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2844/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5319/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6293/)

### DIFF
--- a/rmw_opensplice_cpp/src/demangle.cpp
+++ b/rmw_opensplice_cpp/src/demangle.cpp
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <string>
 #include <algorithm>
+#include <cassert>
+#include <regex>
+#include <string>
 #include <vector>
 
 #include "rcutils/logging_macros.h"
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/demangle.cpp
```diff
@@ -36,19 +38,24 @@ _demangle_if_ros_topic(const std::string & topic_name)
...
-  return dds_type_string;
+
+  // The first submatch is the whole string and the second is the parenthesized expression
+  assert(2u == match.size());
+  std::string substring = match[1].str();
+  size_t substring_position = dds_type_string.find(substring);
+  std::string pkg = dds_type_string.substr(0, substring_position);
+  size_t start = substring_position + substring.size();
+  std::string type_name = dds_type_string.substr(start, dds_type_string.length() - 1 - start);
+  return pkg + "/" + type_name;
```
dirk-thomas: The return value being `<pkg>/<type_name>` isn't going to work anymore - it already looses the namespace. So any code built on top won't be able to map this to the correct type.

jacobperron: Are you worried about existing code or the new action graph API?
For the action graph API, I only wanted to be able to list types associated with the hidden action topics (which works fine with this change).

I think I see what you mean though, existing code would still work, but under the assumption that the type is still in the `msg` namespace, which isn't great.
I can see two solutions:

1. We can add the namespace to the string: `<pkg>/<ns>/<type_name>`
  But I guess this will affect a number of downstream packages.
2. I can assume the returned string will have the namespace `::action::dds_` at the `rcl` layer and do this parsing there.

dirk-thomas: > but under the assumption that the type is still in the msg namespace

That is not the case. Messages are in the same namespace as the type they are derived from:
* the request and response messages of a service are in the `srv` namespace
* the messages and services of an action are in the `action` namespace.

> We can add the namespace to the string: `<pkg>/<ns>/<type_name>`
> But I guess this will affect a number of downstream packages.

It would probably be good to return an actual tuple instead of a string with certain separators (since the logic for splitting will be all over the place). That conceptional change will certainly affect a lot of code / packages.

> I can assume the returned string will have the namespace ::action::dds_ at the rcl layer and do this parsing there.

`rcl` should never see the `dds_` part which is an implementation detail of the typesupport.

jacobperron: In order to return a tuple for the type, we'd need to update the data structure [rmw_names_and_types_t](https://github.com/ros2/rmw/blob/0d7db0fee279a226285e63a594c5448ef4815c57/rmw/include/rmw/names_and_types.h#L32-L37). How about the following structure?

```c
typedef struct RMW_PUBLIC_TYPE rmw_names_and_types_t
{
  rcutils_string_array_t names;
  // The length of this array is the same as names.size
  // Each element of an array list is a type represented by an array list of strings
  // representing the fully qualified type name, e.g. ["pkg_name", "msg", "Empty"]
  rcutils_array_list_t * types;
} rmw_names_and_types_t;
```

If it sounds good to you, I will go ahead and start refactoring the affected packages.

dirk-thomas: Sounds fine with me. I just think this needs a global consensus how we identify interfaces (across various different locations in the code as well as the CLI). Establishing that first (e.g. with a design article) will avoid discussions on how to do it later.

jacobperron: Design PR: ros2/design#220

jacobperron:
Replaced by https://github.com/ros2/rmw_opensplice/pull/268

PR #265, opened 2019-03-26
Add function to get publisher actual qos settings

ivanpauno:
Implements api added in ros2/rmw#169

Connects to ros2/rmw#169


ivanpauno:
> LGTM but for the exact same observations I did on [ros2/rmw_connext#350](https://github.com/ros2/rmw_connext/pull/350).

Done

PR #267, opened 2019-04-30
API updates for preallocation work.

mjcarroll:
Stubs out the expanded API, which isn't used in Opensplice currently

Connects to ros2/rmw#159

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_publisher.cpp
+++ b/rmw_opensplice_cpp/src/rmw_publisher.cpp
@@ -66,7 +66,8 @@ rmw_init_publisher_allocation(
   (void) type_support;
   (void) message_bounds;
   (void) allocation;
-  return RMW_RET_OK;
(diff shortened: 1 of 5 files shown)


rmw_opensplice_cpp/src/rmw_publisher.cpp
```diff
@@ -56,6 +56,27 @@ using rosidl_typesupport_opensplice_cpp::process_topic_name;
 // The extern "C" here enforces that overloading is not used.
 extern "C"
 {
+rmw_ret_t
+rmw_init_publisher_allocation(
+  const rosidl_message_type_support_t * type_support,
+  const rosidl_message_bounds_t * message_bounds,
+  rmw_publisher_allocation_t * allocation)
+{
+  // Unused in current implementation.
```
clalancette: As with the Fast-RTPS one, I think that if the user calls this inadvertently, we shouldn't just pretend that it succeeds.  Same below.

PR #268, opened 2019-05-06
Combine package name with message namespace in type support struct

jacobperron:
Connects to https://github.com/ros2/ros2/issues/677

### DIFF
--- a/rmw_opensplice_cpp/src/demangle.cpp
+++ b/rmw_opensplice_cpp/src/demangle.cpp
@@ -111,7 +111,7 @@ _demangle_service_from_topic(const std::string & topic_name)
 std::string
 _demangle_service_type_only(const std::string & dds_type_name)
 {
-  std::string ns_substring = "::srv::dds_::";
+  std::string ns_substring = "dds_::";
   size_t ns_substring_position = dds_type_name.find(ns_substring);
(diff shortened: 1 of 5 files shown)


rmw_opensplice_cpp/src/demangle.cpp
```diff
@@ -36,19 +37,21 @@ _demangle_if_ros_topic(const std::string & topic_name)
...
+    return dds_type_string;
   }
-  // not a ROS type
-  return dds_type_string;
+
+  std::string type_namespace = dds_type_string.substr(0, substring_position);
+  type_namespace = std::regex_replace(type_namespace, std::regex("::"), "/");
+  size_t start = substring_position + substring.size();
+  std::string type_name = dds_type_string.substr(start, dds_type_string.length() - 1 - start);
+  return type_namespace + type_name;
```
dirk-thomas: Inverting the logic makes the diff more difficult to review. What was the reason to do so?

jacobperron: Consistency. The logic now matches the other `rmw` implementations. I can revert it or break it into two commits if you prefer.

dirk-thomas: No need it - the change looks good. I just wanted to check.

PR #286, opened 2019-10-09
Fix build error introduced in #284

ivanpauno:
See #284


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8316/)

PR #288, opened 2019-10-18
Add localhost parameter to create node function

BMarchi:
Add the parameter to the create node function without implement the logic for it. Connected to [this](https://github.com/ros2/ros2/issues/798)

### DIFF
--- a/rmw_opensplice_cpp/src/rmw_node.cpp
+++ b/rmw_opensplice_cpp/src/rmw_node.cpp
@@ -48,6 +48,7 @@ rmw_create_node(
   const char * name, const char * namespace_, size_t domain_id,
   const rmw_node_security_options_t * security_options, bool localhost_only)
 {
+  static_cast<void>(localhost_only);
   RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, nullptr);
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     init context,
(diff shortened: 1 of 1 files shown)


rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -46,7 +46,7 @@ rmw_node_t *
 rmw_create_node(
   rmw_context_t * context,
   const char * name, const char * namespace_, size_t domain_id,
-  const rmw_node_security_options_t * security_options)
+  const rmw_node_security_options_t * security_options, bool localhost_only)
```
dirk-thomas: The implementation block should contain `(void)localhost_only` to avoid a compiler warning.

rmw_opensplice_cpp/src/rmw_node.cpp
```diff
@@ -48,6 +48,7 @@ rmw_create_node(
   const char * name, const char * namespace_, size_t domain_id,
   const rmw_node_security_options_t * security_options, bool localhost_only)
 {
+  static_cast<void>(localhost_only);
```
hidmic: @BMarchi that's a bit unusual, I'd normally do `(void)var;`.

dirk-thomas:
@eboasson Can you please follow up on this and implement the necessary logic to limit the communication to localhost when the flag is set? Thank you.

hidmic:
@dirk-thomas PTAL again after https://github.com/ros2/rmw_opensplice/pull/288/commits/dfc87a80fa3b48dd510785b8ee2f8ca916b04796.

